PR #3, opened 2017-10-14
Readme updates

oniofchaos:


.travis.yml
```diff
@@ -4,10 +4,9 @@ install: bundle install --jobs=3 --retry=3
 before_script: bundle install
 script: bundle exec rake
 rvm:
-  - 2.1.8
   - 2.2.4
```
olbrich: I'd suggest building against the officially supported ruby versions... that would be 2.4.2, 2.3.5, and 2.2.8

olbrich: see #4, which should cleanup these issues

PR #11, opened 2017-12-05
Remove compound conditionals

jstumbaugh:
Similar to an [earlier PR](https://github.com/devise-security/devise-security/pull/6) on the same subject.

This removes the duplicated checking of the security question's answer. In 3 places, it combines the checking for valid captchas and correct security question answer into a single method check to a new helper method. I think it improves readability a bit.

(This also adds the `valid_captcha_if_defined?` check in 2 places I missed in the earlier PR :see_no_evil: )

### DIFF
--- a/lib/devise-security/controllers/helpers.rb
+++ b/lib/devise-security/controllers/helpers.rb
@@ -26,6 +26,11 @@ module DeviseSecurity
           defined?(valid_captcha?) && valid_captcha?(captcha)
       end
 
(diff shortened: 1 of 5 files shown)


jstumbaugh:
I tried running this test a few times and it looks like its a random failure. I ran it on master with the seed 59950 and it is reproducible on the `master` branch. I'm not too sure why it is failing though.

PR #16, opened 2018-01-22
Deny newer password than n days option

mathieujobin:
this allow to configure devise password_archivable module to deny passwords not only based on counts, but on how old they are.

you can use both options together or separately

cheers,


natebird:
This is interesting. I like it.

What is the reason for the changes to the local time zone instead of UTC?

oniofchaos:
@mathieujobin are you able to fix CI + the merge conflicts?

mathieujobin:
not really, codebase changed too much. sorry

PR #19, opened 2018-01-25
fix filename of generated initializer

maurosbu:


oniofchaos:
Thanks for the cleanup!

PR #20, opened 2018-01-30
Add CircleCI config

natebird:


olbrich:
@natebird what's the motivation for adding circle-ci?  Are you planning to keep travis?

natebird:
@olbrich I find Circle to be much faster and easier to configure. We'll run both for now.

PR #23, opened 2018-02-15
Update Password Archivable to use password_salt

givigier:
We need to add password_salt on migration because we get an error if we don't include this column.


pirj:
Sorry to ask, but what error? And was there a spec added later on to avoid regression of this error?

olbrich:
@pirj looks like this was just a documentation update, not a code change.

PR #25, opened 2018-03-20
Avoid crash if resource model's encrypted_password column is NULL

micapam:
Although by default, Devise [generates](https://github.com/plataformatec/devise/blob/master/lib/generators/active_record/devise_generator.rb) migrations for an `encrypted_password` column that is `NOT NULL`, there are situations in which the column _is_ `NULL`. For instance, the documentation in the `devise_invitable` gem [explicitly requires](https://github.com/scambra/devise_invitable#activerecord-migration) developers using their gem to make that column `NULL`.

So, it's not safe to assume `encrypted_password` is always `NOT NULL`. In the use case where a user is created without a password, and then chooses one, a crash will be caused because `PasswordArchivable` attempts to create an `OldPassword` record with a null `encrypted_password`.

The changes in this PR simply make `PasswordArchivable` check before saving an `OldPassword` to make sure the previous password was not nil (and also introduces an associated test).

### DIFF
--- a/test/dummy/db/migrate/20120508165529_create_tables.rb
+++ b/test/dummy/db/migrate/20120508165529_create_tables.rb
@@ -6,7 +6,7 @@ class CreateTables < MIGRATION_CLASS
 
       ## Database authenticatable
       t.string :email,              null: false, default: ''
-      t.string :encrypted_password, null: false, default: ''
+      t.string :encrypted_password, null: true, default: ''
 
       t.datetime :password_changed_at
       t.timestamps null: false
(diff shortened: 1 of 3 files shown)


test/dummy/db/migrate/20120508165529_create_tables.rb
```diff
@@ -6,7 +6,7 @@ def self.up
 
       ## Database authenticatable
       t.string :email,              null: false, default: ''
-      t.string :encrypted_password, null: false, default: ''
+      t.string :encrypted_password, null: true, default: ''
```
micapam: Not sure if you need this change. I can back it out if not.

natebird: Let's stick with the default (`null: false`) for the migration. Thanks!

micapam: Done :)

PR #26, opened 2018-03-21
Fix deprecation warnings

### LINKED ISSUE #22: RSpec deprecation warning message
When running RSpec tests with the latest version I'm getting deprecation warning messages:

> DEPRECATION WARNING: The behavior of `attribute_changed?` inside of after callbacks will be changing in the next version of Rails. The new return value will reflect the behavior of calling the method after `save` returned (e.g. the opposite of what it returns now). To maintain the current behavior, use `saved_change_to_attribute?` instead.

I'm using the following version
- rails: 5.1.4
- devise: 4.3.0
- devise-security: 0.11.1
- rspec: 3.7.0

olbrich:
This branch does the following:
* adds formalized testing framework for testing against rails 4.2 and 5.1 to help with ensuring backwards compatibility
* refactors the PasswordArchiveable module to deal with changes in the rails dirty attribute handling in rails 5.1
* Cleans up tests a bit
* makes the test harness user include all of the various devise modules so we can check for interactions
* updates supported ruby versions
* adds some debugging tools and configurations to better expose deprecation warnings.

fixes #22


test/dummy/config/environments/test.rb
```diff
@@ -24,4 +24,10 @@
   I18n.enforce_available_locales = false
 
   config.active_support.test_order = :sorted
+  config.log_level = :debug
+  if Rails.version < '5'
+    # TODO: this needs to be addressed
```
natebird: What needs to be addressed here?

olbrich: If I don't set this config in rails 4.2, it will raise a deprecation warning.  I'm not entirely sure there is more to do here or if that warning is just intended to warn people about about a change in behavior.

.gitignore
```diff
@@ -37,3 +37,4 @@ log
 test/tmp/*
 *.gem
 Gemfile.lock
+*.lock
```
oniofchaos: What's this for?

lib/devise-security/models/compatibility.rb
```diff
@@ -0,0 +1,22 @@
+module Devise
+  module Models
+    module Compatibility
```
oniofchaos: Nice!

lib/devise-security/models/password_archivable.rb
```diff
@@ -1,45 +1,47 @@
+require_relative 'compatibility'
+
 module Devise
   module Models
-    # PasswordArchivable
+    # PasswordArchivable, this depends on the DatabaseAuthenticatable module from devise
     module PasswordArchivable
       extend ActiveSupport::Concern
+      include Devise::Models::DatabaseAuthenticatable
```
oniofchaos: Can you alphabetize these?

test/dummy/config/environments/test.rb
oniofchaos: I don't think a TODO here is appropriate, as there's nothing to fix. Instead, a comment explaining why this is needed for older Rails would be good.

olbrich: It was more of a reminder to myself that I forgot to remove.

.gitignore
olbrich: Appraisal generates Gemfiles in the /gemfiles directories.  We don't want to check in their lock files.

natebird:
@olbrich Looks good. Thank you!

I'll just note here that the test coverage only decreased because of the refactoring of the migration code. We should probably exclude some of those files in the future.

Issue #24, opened 2018-03-07
password_regex and ReDoS attacks


micapam:
First of all, thanks for forking the original repo and keeping the functionality alive! 

I'm using devise-security in an app I'm working on. A problem I found is that the `password_regex` setting wasn't usable for my requirements (ensuring passwords include characters from certain classes: e.g. uppercase letter, lowercase letter, a number and a symbol). Whatever regex I tried to accomplish this, I found I was opening the app to a potential ReDos attack:

https://www.owasp.org/index.php/Regular_expression_Denial_of_Service_-_ReDoS

Maybe if someone with better regex skills than me could add some examples of safe regexes to the documentation (and warnings of the issue) that would be helpful. 

Alternatively, I'm using a slightly different way of validating complexity in passwords by specifying required character classes in Devise config. I could raise a PR for this if you like. However, it is necessarily a breaking change, as old configs won't work any more. Perhaps (if you like this method) both could be implemented side by side, with the `password_regex` config deprecated until it's phased out...

natebird:
I'd be happy to review a PR of this feature. Thanks for looking into this.

PR #31, opened 2018-04-05
Password complexity validation with ReDOS vulnerability

micapam:
As I mentioned (in #24), the devise-security gem currently is likely (depending on what password_regex is used) to open up a vulnerability to ReDOS attacks. 

Fixing this cannot be done without changing the means by which password complexity is set - it's the fact that a single regex is used that's the problem. This unavoidably means a breaking change. 

I have implemented a complexity validator using what I believe is the most commonly required POSIX character sets (upper-case letters, lower-case letters, numerals, and punctuation marks/symbols). Client code can specify how many of each they want, or leave out any sets they don't want.

I also added a French translation, which is not great as French is not my native language, but I figured as I knew more French than German or Spanish I may as well add that in too while I was about it :)


lib/devise-security/validators/complexity_validator.rb
```diff
@@ -0,0 +1,23 @@
+module DeviseSecurity
+  # String complexity validator
+  # Validates complexity of e.g. passwords
+  # Options:
+  # - digit: minimum number of digits in the validated string
+  # - lower: minimum number of lower-case letters in the validated string
+  # - punct: minimum number of punctuation characters or symbols in the validated string
```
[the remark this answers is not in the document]
micapam: Agree on the character classes, that makes sense. 

Not sure what you mean about 'sets of repeated characters or sequences', though? Each regex is for a single character so I'm not sure how repeated characters or sequences in the input would cause a problem? 

Another approach would be simply to iterate over each character in the string in turn, bumping the count for any character set it meets (using `unpack('U')` or similar) and do away with regexen altogether. Thoughts?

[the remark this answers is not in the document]
olbrich: Some recommendations for password complexity suggest that you should avoid passwords with long sets of repeated characters (i.e. `aaaaaa`) and sequential characters (`abcdefg` or `1234`).  This is outside the scope of this specific PR as the original code didn't do that either.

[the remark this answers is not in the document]
olbrich: Using `unpack('U')` is an interesting approach.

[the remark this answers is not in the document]
micapam: Oh I _see_. When you said 'protect against', I thought you were talking about protecting against ReDOS attacks (which, incidentally, use repetition, etc).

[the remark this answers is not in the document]
micapam: Just about to head home for the weekend, but I should be able to update this PR on Monday (Australia time - so that's Sunday eve from your perspective :)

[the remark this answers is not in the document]
micapam: And yes, agreed repeated characters would be useful to protect against (but out of scope for this PR, which is just to use a safer method than asking developers to write a regex). Another common request is avoidance of common dictionary words. 

In fact, I don't really think much of the usual complexity rules; I agree with Jeff Atwood that [password rules are bullsh**](https://blog.codinghorror.com/password-rules-are-bullshit/). (And indeed [Bill Burr](https://www.wsj.com/articles/the-man-who-wrote-those-password-rules-has-a-new-tip-n3v-r-m1-d-1502124118).) I'd much rather use a validator like this one: 
https://github.com/cmer/nobspw

However, our clients in their wisdom want the usual kind of complexity rules, based on outdated standards...and they're paying the bills 😂

micapam:
@olbrich, I think I've addressed your feedback regarding character classes. 

Also, as I was adjusting the tests, I found that the `setup` block in `test/test_secure_validatable.rb` did absolutely nothing. The reason is that validators accept their rules when the model is first loaded; they don't look up the config again when it's time to validate. (Incidentally, this is also true for the current master branch, too - the setup block does nothing. Try changing the value for password_regex in there - should break the tests, but doesn't.) 

I considered making the validator accept a Proc, so that configs could be changed at runtime. However, this felt like deforming production code to make testing easier, which wouldn't be good. So instead, I added a unit test for the validator itself, with some dummy models to test various scenarios. See `test/test_complexity_validator.rb`.

test/test_complexity_validator.rb
```diff
@@ -0,0 +1,78 @@
...
+class OneSymbol < Validatable
+  validates :complex_string, 'devise_security/complexity': { lower: 1 }
+end
+
+class OneOfEach < Validatable
+  validates :complex_string, 'devise_security/complexity': { lower: 1, upper: 1, digit: 1, symbol: 1 }
+end
+
+class EmailValidatorTest < Minitest::Test
+  def with_no_rules_anything_goes
```
olbrich: These tests don't execute.  They need to start with `test_`

test/test_complexity_validator.rb
```diff
@@ -0,0 +1,78 @@
...
+    assert(NoRules.new('aaaa').valid?)
+  end
+
+  def enforces_uppercase
```
olbrich: There is another pattern here that may be a bit more explicit.

```
class PasswordComplexityValidatorTest 
  class Password
     include ActiveModel::Validations
     attr_reader :password
     def initialize(password)
       @password = password
    end
    
  def setup
    Password.clear_validators!
  end

  def test_enforces_uppercase
     Password.validates :password, 'devise_security/complexity': { upper: 1 }
     refute(Validateable.new('aaaa').valid?)
     assert(Validateable.new('Aaaa').valid?)
  end 

...
```

lib/devise-security/validators/complexity_validator.rb
```diff
@@ -0,0 +1,28 @@
...
+  # String complexity validator
+  # Validates complexity of e.g. passwords
+  # Options:
+  # - digit:  minimum number of digits in the validated string
+  # - lower:  minimum number of lower-case letters in the validated string
+  # - symbol: minimum number of punctuation characters or symbols in the validated string
+  # - upper:  minimum number of upper-case letters in the validated string
+  class ComplexityValidator < ActiveModel::EachValidator
+    PATTERNS = {
+      digit: /\p{Digit}/,
```
olbrich: I think it's likely that some people will try to use `digits: 2` or `symbols: 2` in the configuration, which is silently fail to do the right thing.  Consider either making them aliases, or checking the keys to make sure they match the set that can be used.

test/test_secure_validatable.rb
```diff
@@ -31,32 +28,32 @@ class User < ActiveRecord::Base
...
+    msg = 'Password must contain at least one lower-case letter'
     user = User.create email: 'bob@microsoft.com', password: 'PASSWORD1', password_confirmation: 'PASSWORD1'
     assert_equal(false, user.valid?)
     assert_equal([msg], user.errors.full_messages)
     assert_raises(ActiveRecord::RecordInvalid) { user.save! }
   end
 
   test 'password must have number' do
-    msg = 'Password must contain big, small letters and digits'
+    msg = 'Password must contain at least one numeral'
```
olbrich: replace `numeral` with `digit` for consistency.

lib/devise-security/validators/complexity_validator.rb
```diff
@@ -0,0 +1,28 @@
+module DeviseSecurity
+  # String complexity validator
+  # Validates complexity of e.g. passwords
+  # Options:
+  # - digit:  minimum number of digits in the validated string
+  # - lower:  minimum number of lower-case letters in the validated string
+  # - symbol: minimum number of punctuation characters or symbols in the validated string
+  # - upper:  minimum number of upper-case letters in the validated string
+  class ComplexityValidator < ActiveModel::EachValidator
```
olbrich: name this `PasswordComplexityValidator`.  You'll also need to rename all the associated tests/files.

olbrich:
👍 I like the approach of testing the validator independently.

micapam:
I think I have addressed all second-round feedback..?

lib/devise-security/models/secure_validatable.rb
```diff
@@ -48,7 +49,7 @@ def self.included(base)
 
           # extra validations
           validates :email, email: email_validation if email_validation # use rails_email_validator or similar
-          validates :password, format: { with: password_regex, message: :password_format }, if: :password_required?
+          validates :password, 'devise_security/password_complexity': password_complexity, if: :password_required?
```
oniofchaos: Can you fix this Rubocop warning while you're in here?

lib/devise-security/validators/password_complexity_validator.rb
```diff
@@ -0,0 +1,29 @@
+module DeviseSecurity
```
oniofchaos: This file has a bunch of Rubocop warnings.

lib/devise-security/validators/password_complexity_validator.rb
```diff
@@ -0,0 +1,29 @@
...
+      digit: /\p{Digit}/,
+      digits: /\p{Digit}/,
+      lower: /\p{Lower}/,
+      upper: /\p{Upper}/,
+      symbol: /\p{Punct}|\p{S}/,
+      symbols: /\p{Punct}|\p{S}/,
+    }.freeze
+
+    def validate_each(record, attribute, value)
+      (options.keys & PATTERNS.keys).each do |key|
```
oniofchaos: Might be able to pull `options.keys & PATTERNS.keys` into a separate method to help with complexity.

lib/devise-security/models/secure_validatable.rb
micapam: Line length? Sure, done, but now CodeClimate is complaining there are too many LoC! 😂

lib/devise-security/validators/password_complexity_validator.rb
micapam: Fixed.

lib/devise-security/validators/password_complexity_validator.rb
micapam: Yep, that worked.

PR #34, opened 2018-04-16
Frozen string literal

oniofchaos:


oniofchaos:
Not gonna fix the snake case issues because that could cause breaking changes elsewhere.

PR #38, opened 2018-04-27
Added Turkish locale

muratbsts:
I was not able to run the tests. 😞

### DIFF
--- a/lib/generators/devise_security/install_generator.rb
+++ b/lib/generators/devise_security/install_generator.rb
@@ -4,7 +4,7 @@ module DeviseSecurity
   module Generators
     # Generator for Rails to create or append to a Devise initializer.
     class InstallGenerator < Rails::Generators::Base
-      LOCALES = %w[ en es de it tr ].freeze
+      LOCALES = %w[en es de fr it tr].freeze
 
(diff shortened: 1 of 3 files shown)


natebird:
As for the failing tests - shouldn't we be using the translation keys in the tests instead of the English strings?

muratbsts:
Hi, @natebird @olbrich I fixed the issues that you said. I ran tests on my machine with ruby v2.4.3 + rails v4.2/5.0/5.2

But travis jobs failed again. ☹️ 

Thank you, I learned new things about of testing.

lib/generators/devise_security/install_generator.rb
```diff
@@ -4,7 +4,7 @@ module DeviseSecurity
   module Generators
     # Generator for Rails to create or append to a Devise initializer.
     class InstallGenerator < Rails::Generators::Base
-      LOCALES = %w[ en de it ]
+      LOCALES = %w[ en es de it tr ].freeze
```
olbrich: please add `fr` to this list before the `it` and the tests will pass.  Also please remove extra spaces at start and end.

muratbsts: 👍 done

Issue #39, opened 2018-05-16
undefined method `migrate' for ActiveRecord::Migrator:Class (NoMethodError)


scicco:
stacktrace is:

`devise-security/test/test_helper.rb:16:in <top (required)>': undefined method 'migrate' for ActiveRecord::Migrator:Class (NoMethodError)
	from /Users/guidoscicolone/prj/devise-security/test/test_captcha_controller.rb:3:in 'require'
	from /Users/guidoscicolone/prj/devise-security/test/test_captcha_controller.rb:3:in '<top (required)>'
	from /Users/guidoscicolone/.rvm/gems/ruby-2.2.9@devise-security/gems/rake-12.3.1/lib/rake/rake_test_loader.rb:17:in 'require'
	from /Users/guidoscicolone/.rvm/gems/ruby-2.2.9@devise-security/gems/rake-12.3.1/lib/rake/rake_test_loader.rb:17:in 'block in <main>'
	from /Users/guidoscicolone/.rvm/gems/ruby-2.2.9@devise-security/gems/rake-12.3.1/lib/rake/rake_test_loader.rb:5:in 'select'
	from /Users/guidoscicolone/.rvm/gems/ruby-2.2.9@devise-security/gems/rake-12.3.1/lib/rake/rake_test_loader.rb:5:in '<main>'
rake aborted!
`

This error appear when trying to lauch test with 

`bundle exec appraisal rake test`

or with 

`bundle exec rake test`

this should be related to the latest changes in Rails migration as already noticed in this [comment](https://github.com/thiagopradi/octopus/issues/477#issue-299041670)

PR #40, opened 2018-05-16
Fix for #39

scicco:
This should fix undefined method `migrate' for ActiveRecord::Migrator:Class #39 issue


scicco:
Hello @natebird I've removed the rc1 version. Thank you for your efforts for this gem.

test/test_helper.rb
```diff
@@ -13,7 +13,12 @@
 
 ActiveRecord::Migration.verbose = false
 ActiveRecord::Base.logger = Logger.new(nil)
-ActiveRecord::Migrator.migrate(File.expand_path('../dummy/db/migrate', __FILE__))
+if Rails.version == '5.2.0'
```
olbrich: What happens with Rails 5.2.1?  This change probably needs to happen for all versions >= 5.2.0

test/test_helper.rb
```diff
@@ -13,7 +13,12 @@
...
+  mc = ActiveRecord::MigrationContext.new(File.expand_path('../dummy/db/migrate', __FILE__))
```
olbrich: not really using `mc`.  Just write it as
```
ActiveRecord::MigrationContext.new(File.expand_path('../dummy/db/migrate', __FILE__)).migrate
```

test/test_helper.rb
scicco: Hello @olbrich you're right, I'll commit the fix.

test/test_helper.rb
scicco: fixed in latest commit

scicco:
Hello @olbrich is the last change ok for you?

test/test_helper.rb
```diff
@@ -13,7 +13,11 @@
 
 ActiveRecord::Migration.verbose = false
 ActiveRecord::Base.logger = Logger.new(nil)
-ActiveRecord::Migrator.migrate(File.expand_path('../dummy/db/migrate', __FILE__))
+if Rails.version >= '5.2.0'
```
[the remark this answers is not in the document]
scicco: fixed, thank you for your review.

olbrich:
@scicco Looks like you will need to update the branch on master once #42 is merged in order for tests to pass for ruby 2.5

PR #42, opened 2018-05-21
Change supported ruby version from 2.5.0 to 2.5.1 (fixes #41)

### LINKED ISSUE #41: Add support for ruby 2.5.1, drop 2.5.0
Ruby 2.5.0 seems to cause problems installing bundler on Travis, so drop it in favor of ruby 2.5.1

olbrich:
Travis seems to have problems running with ruby 2.5.0 so we switch support to 2.5.1


olbrich:
@natebird tests still failed, I'm trying to troubleshoot this.  It seems to be more complicated than just an older ruby version.

olbrich:
Looks like it might have actually been due to a recent bundler release... https://github.com/bundler/bundler/releases 
I can't find something in there that would specifically break it, but I think it has to do with resolving binstubs.

Issue #36, opened 2018-04-17
Support Mongoid


natebird:
Is it worth supporting Mongoid? Devise supports it.

oniofchaos:
I personally don't think so. At best, it would be in a separate gem.

olbrich:
I don't have a sense of how much people even use mongoid.  I might be willing to investigate it if a lot of users expressed an interest.

scicco:
+1 
We are currently using Mongoid with devise and adding devise-security features into our application will help us a lot to be compliant with GDPR requirements.

brunoporto:
https://github.com/devise-security/devise-security/pull/43

DBDeveloper87:
I am beginning my app development on Active Record ORM, but am also learning how to work with MongoDB at the same time.  I do not yet understand MongoDB enough to use it properly, however its capabilities are in line with plans for future development of my application, so I plan to switch ORMs then.
For now, I'll use Devise-Security; however, when I make the switch to mongoID if this gem doesn't yet support it, I will need some alternative for the features you have in your gem.

scicco:
Please take a look at #76 . If someone can review my work and help fixing the failing test.

olbrich:
@scicco what version of Devise and Mongoid are you using?

scicco:
Hello @olbrich, this is taken from my old Gemfile.lock (I didn't pull the latest changes)

```
  ...
    mongoid (7.0.2)
      activemodel (>= 5.1, < 6.0.0)
      mongo (>= 2.5.1, < 3.0.0)
  ...
    devise (4.6.1)
      bcrypt (~> 3.0)
      orm_adapter (~> 0.1)
      railties (>= 4.1.0, < 6.0)
      responders
      warden (~> 1.2.3)
  ...
```

PR #43, opened 2018-06-21
 mongoid compatibility

brunoporto:
mongoid compatibility
https://github.com/devise-security/devise-security/issues/36

### DIFF
--- a/lib/devise-security/models/expirable.rb
+++ b/lib/devise-security/models/expirable.rb
@@ -22,7 +22,8 @@ module Devise
 
       # Updates +last_activity_at+, called from a Warden::Manager.after_set_user hook.
       def update_last_activity!
-        self.update_column(:last_activity_at, Time.now.utc)
+        self.last_activity_at = Time.now.utc
+        self.save
       end
 
       # Tells if the account has expired
--- a/lib/devise-security/models/old_password.rb
+++ b/lib/devise-security/models/old_password.rb
@@ -1,6 +1,0 @@
-# frozen_string_literal: true
-
-require 'active_record'
-class OldPassword < ActiveRecord::Base
-  belongs_to :password_archivable, polymorphic: true
-end
--- a/README.md
+++ b/README.md
@@ -158,6 +158,14 @@ add_index :the_resources, :password_changed_at
 ```
 
 ### Password archivable
+
+`app/models/old_password.rb`
+**ActiveRecord**
+```ruby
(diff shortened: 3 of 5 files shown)


lib/devise-security/models/old_password.rb
```diff
@@ -1,6 +0,0 @@
-# frozen_string_literal: true
-
-require 'active_record'
-class OldPassword < ActiveRecord::Base
```
oniofchaos: I don't understand why this is removed.

lib/devise-security/models/expirable.rb
```diff
@@ -22,7 +22,8 @@ module Expirable
 
       # Updates +last_activity_at+, called from a Warden::Manager.after_set_user hook.
       def update_last_activity!
-        self.update_column(:last_activity_at, Time.now.utc)
+        self.last_activity_at = Time.now.utc
```
oniofchaos: Fix the CC issue here

lib/devise-security/models/old_password.rb
brunoporto: Moved to app, because i can't verify mongoid/activerecord after inheritance, but I can verify and generate right model structure in the installation

oniofchaos: I'll let the other maintainers weigh in but that doesn't fly with me. It makes it much harder for users to use this gem out of the box and moves this model out of our control if we want to make changes. We've discussed splitting this gems out into smaller gems and maybe after that point there can be  `password-archivable` and `password-archivable-mongoid` as separate gems.

brunoporto: I agree, much better.

Personally I prefer the activerecord but where I work they use mongoid and I was forced to create this portability, then I decided to do the PR to collaborate with others in my situation but I agree with you that having to create hacks in the code for this is bad.

lib/devise-security/models/expirable.rb
olbrich: There is a subtle difference here in behavior.  The original implementation  bypasses validation checks, while this approach does not.  It's possible for this call to fail if the record is in an invalid state for some reason.

oniofchaos:
We very much appreciate your contribution. Unfortunately, we're not going to accept this PR as-is and are going to close it. Feel free to continue discussion on the linked issue if you (or others) are interested in adding Mongo capability to this gem.

Issue #44, opened 2018-06-21
Fix Deprecation warning -- Support Rails >= 5


SauloSilva:
DEPRECATION WARNING: The behavior of `attribute_changed?` inside of after callbacks will be changing in the next version of Rails. The new return value will reflect the behavior of calling the method after `save` returned (e.g. the opposite of what it returns now). To maintain the current behavior, use `saved_change_to_attribute?` instead.

This warning is [here](https://github.com/devise-security/devise-security/blob/master/lib/devise-security/models/password_expirable.rb#L53).

olbrich:
@SauloSilva How did you get that deprecation warning?  There is only a `before_save` callback in that code, not the `after_save` callback the message describes.

PR #45, opened 2018-06-23
Refactor Password Expireable

olbrich:
Mostly this is a cleanup of the password expireable code that was started in part because of #44.  I haven't yet been able to replicate the original issue, but it seemed like a good excuse to clean up some older code.


lib/devise-security/models/password_expirable.rb
```diff
@@ -4,66 +4,83 @@
...
-          return @_numeric if defined?(@_numeric)
-          @_numeric ||= self.expire_password_after.is_a?(1.class) ||
-            self.expire_password_after.is_a?(Float)
-        end
+      # Enabled if configuration +expire_password_after+ is set to an {Integer}, {Float}, or {true}
+      def enabled?
+        return @enabled if defined?(@enabled)
+        @enabled = expire_password_after.is_a?(1.class) ||
+                   expire_password_after.is_a?(Float) ||
+                   expire_password_after == true
```
oniofchaos: Seems like you can use `expire_on_demand?` here

lib/devise-security/models/password_expirable.rb
```diff
@@ -4,66 +4,83 @@
...
-        if expired_password_after_numeric?
-          self.password_changed_at = self.expire_password_after.seconds.ago
-        end
-
-        # is date not set it will set default to need set new password next login
-        need_change_password if self.password_changed_at.nil?
-
-        self.password_changed_at
+        return unless enabled?
+        self.password_changed_at = nil
```
oniofchaos: Seems like this is a logic change from the original, not just cleaning up CC. Is that intentional?

lib/devise-security/models/password_expirable.rb
```diff
@@ -4,66 +4,83 @@
...
-        if expired_password_after_numeric?
-          self.password_changed_at.nil? || self.password_changed_at < self.expire_password_after.seconds.ago
+        if enabled?
+          if expire_on_demand?
+            # Only expire when password_changed_at is nil
+            password_changed_at.nil?
+          else
+            # Expire when password_changed_at is nil or when the last time the
+            # password was changed is sufficiently old
+            password_changed_at.nil? || password_changed_at < expire_password_after.seconds.ago
```
oniofchaos: Possibly extract this line into a private method to address the CC issue? It's borderline to me.

lib/devise-security/models/password_expirable.rb
olbrich: This is a slightly different implementation.  The original would make up a fake time that would cause the record to appear as if it needed to be changed.  I didn't particularly care for that because it makes it look like the user changed their password at a time when they really didn't and it makes it impossible to distinguish between an administrative expiration of the password and a real one.  Just setting the `password_changed_at` to `nil` has the same effect and it's possible to tell that the password was deliberately expired because that is the only way to get a `nil` in that column.

bjensen:
+1 on someone to review

.rubocop.yml
```diff
@@ -1,5 +1,5 @@
 AllCops:
-  TargetRubyVersion: 2.3
+  TargetRubyVersion: 2.2
```
oniofchaos: Why the switch here?

README.md
```diff
@@ -101,30 +101,37 @@ end
...
+
 ```erb
 <p><%= captcha_tag %></p>
 <p><%= text_field_tag :captcha %></p>
 ```
 
 ## Schema
 
-Note: Unlike Devise, devise-security does not currently support mongoid.  Pull requests are welcome!
+Note: Unlike Devise, devise-security does not currently support mongoid. Pull requests are welcome!
```
oniofchaos: I'd like to remove the `Pull requests are welcome!` given that we've discussed on an issue recently about that being pretty complicated.

lib/devise-security/models/password_expirable.rb
```diff
@@ -2,68 +2,101 @@
...
+    # password update.
+    # @return [void]
+    def need_change_password
+      return unless password_expiration_enabled?
+      self.password_changed_at = nil
+    end
+    alias expire_password need_change_password
+    alias request_password_change need_change_password
+
+    # Set this value to a number of seconds to have passwords expire after a time
```
oniofchaos: This comment is a little confusing because it's not a setter.

lib/devise-security/models/password_expirable.rb
```diff
@@ -2,68 +2,101 @@
...
+    # Set to +true+ if you want to be able to manually expire passwords on demand
+    # without a time limit.
+    # @return [Numeric, Boolean]
+    def expire_password_after
+      self.class.expire_password_after
```
oniofchaos: Is this something you can use delegate for?

lib/devise-security/models/password_expirable.rb
```diff
@@ -2,68 +2,101 @@
...
+    end
 
-        def expired_password_after_numeric?
-          return @_numeric if defined?(@_numeric)
-          @_numeric ||= self.expire_password_after.is_a?(1.class) ||
-            self.expire_password_after.is_a?(Float)
-        end
+    # Enabled if configuration +expire_password_after+ is set to an {Integer}, {Float}, or {true}
+    def password_expiration_enabled?
+      expire_password_after.present? &&
```
oniofchaos: `` expire_password_after.present?` is already part of `expire_password_on_demand?`

test/test_helper.rb
```diff
@@ -3,13 +3,24 @@
 ENV['RAILS_ENV'] ||= 'test'
 
 require 'simplecov'
-require 'coveralls'
-Coveralls.wear!
+SimpleCov.start do
+  add_filter 'gemfiles'
+  add_filter 'test'
```
oniofchaos: If possible I like to only restrict certain files vs all of `test`, because simplecov will catch dead test code.

.rubocop.yml
olbrich: we claim to support 2.2, but rubocop was looking for 2.3 syntax.

README.md
olbrich: This is just a reformatting, if we want to remove that line, I suggest a new PR.

.codeclimate.yml
```diff
@@ -1,37 +1,63 @@
...
+- "dist/"
+- "features/"
+- "**/node_modules/"
+- "script/"
+- "**/spec/"
+- "**/test/"
+- "**/tests/"
+- "**/vendor/"
+- "**/*.d.ts"
+- "gemfiles/"
```
oniofchaos: EOL issue?

.gitignore
```diff
@@ -38,4 +38,5 @@ test/tmp/*
 *.gem
 Gemfile.lock
 *.lock
-bin/
+bin/*
+.yardoc
```
oniofchaos: EOL issue?

PR #46, opened 2018-08-17
Documentation for initializer config file improved

viniciusoyama:
Hi!

I think that my change regarding the `deny_old_passwords` configuration paramenter can help other people when configuring this gem.

A member of my team got confused with the old documentation and though that the value should be a `:count` symbol and not the number of old passwords to be tested :P .


oniofchaos:
Thanks!

PR #54, opened 2018-10-13
Remove unnecessary locks from development dependencies

oniofchaos:


devise-security.gemspec
```diff
@@ -30,15 +30,15 @@ Gem::Specification.new do |s|
...
   s.add_development_dependency 'm'
   s.add_development_dependency 'minitest'
   s.add_development_dependency 'pry-byebug'
   s.add_development_dependency 'pry-rescue'
   s.add_development_dependency 'pry'
-  s.add_development_dependency 'rails_email_validator', '~> 0'
-  s.add_development_dependency 'rubocop', '~> 0.58.0'
-  s.add_development_dependency 'sqlite3', '~> 1.3', '>= 1.3.10'
+  s.add_development_dependency 'rails_email_validator'
+  s.add_development_dependency 'rubocop'
```
olbrich: rubocop needs to be in sync with the version used in codeclimate or we may get conflicting or inconsistent corrections.

oniofchaos: I've always used the latest version at work and haven't had a problem with codeclimate being out of sync in practice. Thoughts on leaving as-is?

oniofchaos: Took this change out for now

PR #55, opened 2018-10-13
Remove Ruby 2.2 support

oniofchaos:
https://www.ruby-lang.org/en/news/2018/06/20/support-of-ruby-2-2-has-ended/


oniofchaos:
Not sure what update you mean - looks like the target version is currently 2.3, which matches up with the update I'm making.

olbrich:
https://github.com/devise-security/devise-security/blob/96bfdd894967978f597504dbf7d5e0d8a5a836a1/.rubocop.yml#L2
should be `2.3`

oniofchaos:
Weird....on my local machine it was already showing as 2.3. I fixed it through Github.

PR #56, opened 2018-10-13
Drop support for Rails 4.1 per Rails maintenance policy

oniofchaos:
https://guides.rubyonrails.org/maintenance_policy.html


oniofchaos:
Updated the docs + updated `gemspec` to require `4.2` instead of `4.1` as that's what I should have done in the first place...

PR #58, opened 2018-10-18
Use latest rubocop on both gemspec and CC

oniofchaos:
We were already using this in Gemfile.lock


devise-security.gemspec
```diff
@@ -39,6 +39,6 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'pry-rescue'
   s.add_development_dependency 'pry'
   s.add_development_dependency 'rails_email_validator'
-  s.add_development_dependency 'rubocop', '~> 0.58.0'
+  s.add_development_dependency 'rubocop'
```
olbrich: I actually think we should lock this to the same version as the codeclimate channel to ensure they stay in sync.

Issue #62, opened 2018-10-26
How to make config dynamic


timsatterfield:
Wondering how to best make the config attributes dynamic per instance of a model.

Example: for a given Model which belongs to a specific Organization might have different password validation requirements or password expiry length.

Overriding password_complexity, password_length or other methods in the Model doesn't seem to do anything. Any help here?

olbrich:
Can you give me a made up code example of what you would like to be able to do?

timsatterfield:
```ruby
class Manager < ActiveRecord::Base
  devise :database_authenticatable, :registerable,
         :recoverable, :trackable,
         :confirmable, :lockable,
         :secure_validatable

  belongs_to :team

  # Override password_length enforced by secure_validatable
  def password_length
  	team.password_length | # return value from the devise config file
  end

  def password_complexity
  	team.password_complexity | # return value from the devise config file
  end

  # Have similar overrides for other module rules etc

end
```

olbrich:
On the surface this seems like it could work, but I suspect what is happening is that those methods are being evaluated at the time the class is loaded and never again after that.  I'll dig into it some more and see if we can verify this and modify this so that it can be a proc or something.

rmorlok:
Is there any way that something like this could make its way into the gem? I have a similar requirement where I have customers with sets of users that have differing IT policies on things like password complexity requirements, etc.

olbrich:
@rmorlok yes, I'll update #63 and see if we can get this working again.

AndrewJGregory:
Any updates on this?

khan-arkhitech-2020:
Hey guyz whats the status for this task

WilfredRuck:
Hey, I could also make use of this. Any updates or in between solutions available?

dillonwelch:
This will be released in `0.17.0` shortly.

PR #63, opened 2018-10-27
Make it possible to override validation configs by overriding certain methods

olbrich:
See issue #62


olbrich:
Bah! may only work with Rails 5.2... 😞

gsar:
@olbrich thanks for your work on this, would love to get it to the point of being merged and released. if you can resolve the merge conflicts, i can help test it.

lib/devise-security/validators/password_complexity_validator.rb
```diff
@@ -19,17 +19,29 @@ class DeviseSecurity::PasswordComplexityValidator < ActiveModel::EachValidator
...
         record.errors.add attribute, :"password_complexity.#{key}", count: minimum
       end
     end
   end
 
-  def active_pattern_keys
-    options.keys & PATTERNS.keys
+  def constraints(record)
+    # TODO - there are examples where there is and isn't a `constraints`. Is one format deprecated?
+    constraints = options[:constraints] || options
```
tmr08c, line 34: Based on the other test changes, it looks like maybe you were moving towards requiring the `constraints` key. Would this be a breaking change?

lib/devise-security/models/secure_validatable.rb
```diff
@@ -5,73 +5,114 @@
...
-          # validate login in a strict way if not yet validated
-          unless has_uniqueness_validation_of_login?
-            validation_condition = "#{login_attribute}_changed?".to_sym
+        already_validated_email = false
 
+        # login attribute should be unique
+        unless has_uniqueness_validation_of_login?
+          # Only perform uniqueness check when the login attribute is dirty so
+          # we can avoid hitting the database unnecessarily.
+          if defined?(will_save_change_to_attribute?)
```
tmr08c, line 57: This was a pretty hacky-fix to get a test passing. Not sure if we should use an older form of dirty checking in the `else` block.

lib/devise-security/models/password_archivable.rb
```diff
@@ -5,13 +5,13 @@
 
 module Devise
   module Models
-    # PasswordArchivable, this depends on the DatabaseAuthenticatable module from devise
```
oniofchaos, line 8: I think the existing is better as the module is still called `DatabaseAuthenticatable` even though the import statement looks different.

lib/devise-security/models/secure_validatable.rb
oniofchaos: This is also needed for Rails updates IIRC as they changed the method name between recent versions.

oniofchaos: Yup looks like my memory is correct, see line 97 and line 98.

lib/devise-security/models/secure_validatable.rb
```diff
@@ -5,93 +5,208 @@
...
-          already_validated_email = false
+        devise :database_authenticatable
 
-          # validate login in a strict way if not yet validated
-          unless has_uniqueness_validation_of_login?
-            validation_condition = "#{login_attribute}_changed?".to_sym
+        already_validated_email = false
 
+        # login attribute should be unique
+        unless has_uniqueness_validation_of_login?
```
oniofchaos, line 89: For code org purposes, can we make this block of code its own method?

lib/devise-security/models/secure_validatable.rb
```diff
@@ -5,103 +5,218 @@
...
         true
       end
 
-      module ClassMethods
+      # Password complexity configurations for an instance of the base class,
+      # defaults to the class method. Override this for more fine-grained
+      # control of complexity configurations based on instance attributes.
+      #
+      # @return [Hash]
+      def password_complexity
```
tmr08c, line 172: Would an `@abstract` tag make sense? 

Screen Shot 2020-09-28 at 7 12 42 PM

I feel like it helps make it more obvious it can be overwritten, but I think the expectation may be that `@abstract` does not give a default.

tmr08c (commented):
Incomplete review (got a bit distracted with Yarddocs), but wanted to share what I hard so far. 

Loving all the docs! 

🐱  Curious, did you explore the idea of [`class_attributes`](https://apidock.com/rails/Class/class_attribute). I haven't thought it through, so I'm not quite sure if it would work, but it seems like it would give class and instance methods, and handle sub-classing well.

lib/devise-security/models/secure_validatable.rb
tmr08c, line 172: Would it make sense to make these (the overridable instance methods), `protected`? This could help prevent exposing these methods for classes that use this module.

lib/devise-security/validators/password_complexity_validator.rb
```diff
@@ -1,35 +1,47 @@
...
-# - upper:  minimum number of upper-case letters in the validated string
+# This validator class is used to determine if a password complies with the
+# specified complexity requirements. For example:
+#
+#     validates_with DeviseSecurity::PasswordComplexityValidator,
+#       attributes: :password,
+#       digits: 1, # minimum number of digits in the validated string (also :digit)
+#       lower: 1, # minimum number of lower-case letters in the validated string
+#       symbols: 1, # minimum number of punctuation characters or symbols in the validated string (also :symbol)
+#       upper: 1, # minimum number of upper-case letters in the validated string
```
tmr08c, line 11: Suggestion to use `@example` tag.

```suggestion
# specified complexity requirements.
#
# @example Validate password has one of each possible type of character
#     attributes: :password,
#     digits: 1, # minimum number of digits in the validated string (also :digit)
#     lower: 1, # minimum number of lower-case letters in the validated string
#     symbols: 1, # minimum number of punctuation characters or symbols in the validated string (also :symbol)
#     upper: 1, # minimum number of upper-case letters in the validated string
#
```

Before:

Screen Shot 2020-10-04 at 3 11 43 PM

After:

Screen Shot 2020-10-04 at 3 11 33 PM

lib/devise-security/models/secure_validatable.rb
```diff
@@ -5,103 +5,218 @@
...
+      #
+      # @return [Range]
+      def password_length
+        self.class.password_length
+      end
+
+      # Email validation options. Override this at the class or instance level
+      # to override global configurations.
+      #
+      # @return [Hash] email validation options
```
tmr08c, line 187:
```suggestion
      # @return [Boolean] whether email validation is enabled
```

Based on the default set up in `devise-security` it looks like this should be a Boolean:

https://github.com/devise-security/devise-security/blob/d11e44a7f496d88a66f6f82513a7b887ffa4ef01/lib/devise-security.rb#L29-L32

It also looks like we may not have added any tests for this method.

tmr08c (commented):
A few small suggestions and mostly stilly style/docs comments. I think the biggest thing that stuck out was around `email_validation`. 

I started thinking about [`class_attributes`](https://apidock.com/rails/Class/class_attribute) again and think they could be an interesting fit, but it seems like this may not play nice with the current setup of `mattr_accessor`, so that's definitely out of scope for this change.

olbrich:
Superseded by #356.  There are some documentation and other changes that will show up in other PRs.

PR #68, opened 2018-12-13
Remove mass-assignment from :password_archivable

npalrecha:
You are required to use `attr_accessible` in order to instantiate and set `encrypted_password` in the same line using `.new`. `encrypted_password` shouldn't be mass-assignable. This change removes the necessity of `encrypted_password` to be mass-assignable.


olbrich:
Please add a description/explanation for why this change is necessary / useful.

shaicoleman:
This fixes a bug where devise-security skips the password history check in certain cases,
e.g. using the protected_attributes_continued gem and not having :encrypted_password in attr_accessible.

There are other instances in the codebase where there's mass assignment, but this is a start

oniofchaos:
Can you add a regression test?

PR #74, opened 2019-02-11
Add ability to skip session limitable on a per record basis

surjay:
This just adds the ability to skip session limitable on a per record or instance basis. Our app has had this forked from the old `devise_security_extensions` gem for a long time and as we're switching over to `devise-security` so that we can upgrade rails it felt like a good time to see about getting this merged in.

The use case for our app is this is a client/account setting, so we need some way to bypass this based on a record instance's logic.


test/test_session_limitable.rb
```diff
@@ -0,0 +1,18 @@
...
+  class ModifiedUser < User
+    def skip_session_limitable?
+      true
+    end
+  end
+
+  test 'record instances can override default check' do
+    user = User.create email: 'bob@microsoft.com', password: 'password1', password_confirmation: 'password1'
+    assert_equal(false, user.skip_session_limitable?)
+    modified_user = ModifiedUser.create email: 'bob2@microsoft.com', password: 'password1', password_confirmation: 'password1'
```
oniofchaos: Can you make this in a separate test? That way we only have one assertion per test, one for default state of false and one for the modified state of true.

surjay: Absolutely

surjay: Split them out.

kirichkov:
@surjay Can you add some documentation on how your session limitable bypass feature works? I'd appreciate it if we can find how to integrate our versions of session limitable to work together.

surjay:
> @surjay Can you add some documentation on how your session limitable bypass feature works?
> 
> I'd appreciate it if we can find how to integrate our versions of session limitable to work together. #79

Sure thing. At a basic level the idea is a record or instance can override the check by defining that method. For example, you could have the following model:

```
class User < ActiveRecord::Base
  devise :session_limitable
end

class ModifiedUser < User
  def skip_session_limitable?
    true
  end
end
```

`ModifiedUser` instances would then skip the session limitable check.  That's a very basic example, the actual use case we have is something like this:

```
class User < ActiveRecord::Base
  devise :session_limitable
  
  belongs_to :account 

  def skip_session_limitable?
    # lots of logic based on account settings and subscriptions, etc. for external users
    # bypass for internal/support users, etc
  end
end
```

kirichkov:
My use case is to "impersonate" users to provide support, without requiring the user's password or resetting their password.

My impression is that your implementation requires persisting the `skip_session_limitable?` flag/logic in the database, which is not applicable in my use case, as it'll allow the user themselves to log multiple times. Is my understanding correct?

surjay:
> My use case is to "impersonate" users to provide support, without requiring the user's password or resetting their password.
> 
> My impression is that your implementation requires persisting the `skip_session_limitable?` flag/logic in the database, which is not applicable in my use case, as it'll allow the user themselves to log multiple times. Is my understanding correct?

My case _could_ require persistence. The method override could really contain any kind of logic you want so its left more up to the application logic to determine how to handle that.  It could be a db field, it could just be model overrides or a config setting.

It could also possibly work with a virtual `attr_accessor` field on the model or something similar where you could theoretically do something like:

```
class User
  attr_accessor :skip_session_check 

  def skip_session_limitable?
    skip_session_check.present?
  end
end

# then in a controller somewhere
current_user.skip_session_limitable = true
sign_in(current_user) # or set_user or whatever you have going on.
```

I think that'd probably work as well.  There is also the bypass devise flag when signing in:

`sign_in(current_user, bypass: true)`

That'll skip warden callbacks. Perhaps that could be another way for you to 'impersonate' a user.

kirichkov:
> My case _could_ require persistence. The method override could really contain any kind of logic you want so its left more up to the application logic to determine how to handle that. It could be a db field, it could just be model overrides or a config setting.

Due to reloading the user from the database on each request by warden, it does require some sort of persistence if you need per-session bypass. The model by design doesn't have access to the session.

> It could also possibly work with a virtual `attr_accessor` field on the model or something similar where you could theoretically do something like:
> 
> ```
> class User
>   attr_accessor :skip_session_check 
> 
>   def skip_session_limitable?
>     skip_session_check.present?
>   end
> end
> 
> # then in a controller somewhere
> current_user.skip_session_check = true
> sign_in(current_user) # or set_user or whatever you have going on.
> ```
> I think that'd probably work as well. There is also the bypass devise flag when signing in:
> 
> `sign_in(current_user, bypass: true)`
> 
> That'll skip warden callbacks. Perhaps that could be another way for you to 'impersonate' a user.

I attempted implementing it that way and it doesn't really work, because it bypasses all warden callbacks, not just the session limitable one, hence why I'm in favor setting this via an option from the controller. After all devise is supposed to do user session authentication, rather than authorization. Authorization makes sense to be implemented at the model/user instance level.

@oniofchaos any input on your side, as a maintainer? I think this has become more of a philosophical question really. I also think both patches could be accepted with a simple `skip = record.skip_session_limitable? || options[:skip_session_limitable]` in the non-fetch callback and `skip = record.skip_session_limitable? || warden.session(options[:scope])['devise.skip_session_limitable']` in the `fetch` callback.

so whichever option is present the session limitable is skipped. If we can agree on such an implementation I'd be happy to rebase my changes on top of @surjay's changes and allow the possibility to skip session limitations based on either model-attached flag or a session/controller attached flag.

oniofchaos:
> also think both patches could be accepted with a simple `skip = record.skip_session_limitable? || options[:skip_session_limitable]` in the non-fetch callback and `skip = record.skip_session_limitable? || warden.session(options[:scope])['devise.skip_session_limitable']` in the `fetch` callback.

I agree with this approach. I can see where both could be useful in different scenarios. @olbrich @natebird thoughts?

surjay:
Any consensus on this?

rusllonrails:
Any updates on this guys?

olbrich:
@surjay We need to rebase this on the recent updates to session limitable

olbrich:
@surjay I updated the underlying implementation of session limitable in master, let's take a look at this again now that it's been rebased and see if we can get this working.

kirichkov:
Once you get @surjay PR in, I'll rebase on top. Hopefully we can get both functions inside the same release.

surjay:
> @surjay @kirichkov I updated the underlying implementation of session limitable in master, let's take a look at this again now that it's been rebased and see if we can get this working.

Sounds good, are we wanting me to make additional changes or what?

olbrich:
@surjay make sure this branch still does what you need with the latest changes.

surjay:
> @surjay make sure this branch still does what you need with the latest changes.

Preliminary testing looks good. I don't see anything in your changes that would affect this small addition.

oniofchaos:
@surjay can you merge in latest master?

surjay:
> @surjay can you merge in latest master?

Yep, done.

PR #76, opened 2019-02-13
Mongoid support

scicco:
This pull is trying to address #36.
I looked inside devise gem to try to understand a proper way to organize the testing part for Activerecord and Mongoid.
I still have a problem with a failing test (for both ORM).

To setup your ORM you have to provide DEVISE_ORM env variable

`DEVISE_ORM=active_record` (default)
or
`DEVISE_ORM=mongoid`


lib/devise-security/models/expirable.rb
```diff
@@ -22,7 +22,11 @@ module Expirable
 
       # Updates +last_activity_at+, called from a Warden::Manager.after_set_user hook.
       def update_last_activity!
-        self.update_column(:last_activity_at, Time.now.utc)
+        if defined? ActiveRecord
```
oniofchaos: IMO a better check would be to see if `update_column` is defined.

lib/devise-security.rb
```diff
@@ -1,6 +1,34 @@
 # frozen_string_literal: true
 
-require 'active_record'
+unless defined?(DEVISE_ORM)
+  DEVISE_ORM = (ENV["DEVISE_ORM"] || :active_record).to_sym
+  DEVISE_ORM = :mongoid
```
oniofchaos: This looks like it's always going to set the constant to `:mongoid`. What I think you want is `DEVISE_ORM ||= :mongoid`

lib/devise-security/models/old_password.rb
```diff
@@ -1,6 +1,28 @@
 # frozen_string_literal: true
 
-require 'active_record'
-class OldPassword < ActiveRecord::Base
-  belongs_to :password_archivable, polymorphic: true
+if DEVISE_ORM == :active_record
```
oniofchaos: IMO instead of doing this we should have two files, one for the AR version and one for the Mongoid version. We then `require` the appropriate one based on this value

lib/devise-security/models/password_archivable.rb
```diff
@@ -39,7 +39,7 @@ def max_old_passwords
       # @return [false] if disabled or not previously used
       def password_archive_included?
         return false unless max_old_passwords > 0
-        old_passwords_including_cur_change = old_passwords.order(:id).reverse_order.limit(max_old_passwords).pluck(:encrypted_password)
+        old_passwords_including_cur_change = old_passwords.order(created_at: :desc).limit(max_old_passwords).pluck(:encrypted_password)
```
oniofchaos: Is this related to Mongoid?

lib/devise-security/models/secure_validatable.rb
```diff
@@ -90,7 +90,9 @@ module ClassMethods
 
         def has_uniqueness_validation_of_login?
           validators.any? do |validator|
-            validator.kind_of?(ActiveRecord::Validations::UniquenessValidator) &&
+            defined? ActiveRecord && validator.kind_of?(ActiveRecord::Validations::UniquenessValidator) &&
+              validator.attributes.include?(login_attribute)
```
oniofchaos: You should have a `next` here, because it's possible that a project could have both `ActiveRecord` and `Mongoid` loaded.

test/dummy/config/mongoid.yml
```diff
@@ -0,0 +1,10 @@
+test:
+  clients:
+    default:
+      database: devise_security_test
+      hosts:
+        - localhost:27017
+      options:
+        read:
+          mode: :primary
+        max_pool_size: 1
```
oniofchaos: EOF issue

.gitignore
```diff
@@ -40,3 +40,5 @@ Gemfile.lock
 *.lock
 bin/*
 .yardoc
+
+.idea
```
oniofchaos: EOF issue

lib/devise-security/models/expirable.rb
scicco: Do you prefer having a check like this?

I have to leave the elsif check because both active_record and mongoid have update_attribute method.

```
      # Updates +last_activity_at+, called from a Warden::Manager.after_set_user hook.
      def update_last_activity!
        if respond_to?(:update_column)
          self.update_column(:last_activity_at, Time.now.utc)
        elsif defined? Mongoid
          self.update_attribute(:last_activity_at, Time.now.utc)
        end
      end
```

lib/devise-security.rb
scicco: I'll remove that line. It's old code, when I was only testing mongoid ORM.

lib/devise-security/models/secure_validatable.rb
scicco: a cleaner version could be:

```
          validators.any? do |validator|
            validator_orm_klass = defined?(ActiveRecord) ? ActiveRecord::Validations::UniquenessValidator : ::Mongoid::Validatable::UniquenessValidator
            validator.kind_of?(validator_orm_klass) && validator.attributes.include?(login_attribute)
          end
```

but I think it shouldn't work in your example. Can you help me with this?

oniofchaos: I just mean that you need something like `next if validator.attributes.include?(login_attribute)`.

lib/devise-security/models/password_archivable.rb
scicco: yes. This way it works for both ORMs

.gitignore
scicco: fixed

.travis.yml
```diff
@@ -9,6 +9,141 @@ rvm:
   - 2.6
   - ruby-head
 matrix:
+  include:
```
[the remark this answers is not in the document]
scicco: ok, done

lib/devise-security.rb
```diff
@@ -1,6 +1,34 @@
 # frozen_string_literal: true
 
-require 'active_record'
+unless defined?(DEVISE_ORM)
+  DEVISE_ORM = (ENV["DEVISE_ORM"] || :active_record).to_sym
```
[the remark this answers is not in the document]
scicco: ok, I'll change like that. Thank you.

lib/devise-security.rb
```diff
@@ -1,6 +1,34 @@
...
+  #       puts '*' * 80
+  #       puts ''
+  #     end
+  #   end
+  #   #print "Requires #{s}\n" if orig_require(s)
+  #   orig_require(s)
+  # end
+
+  require 'mongoid'
+  Mongoid.load!(File.expand_path('../../test/dummy/config/mongoid.yml', __FILE__))
```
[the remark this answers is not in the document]
scicco: ok

lib/devise-security/models/expirable.rb
scicco: done

test/dummy/config/mongoid.yml
scicco: now it should be fixed

.travis.yml
```diff
@@ -18,3 +153,7 @@ gemfile:
   - gemfiles/rails_5.0_stable.gemfile
   - gemfiles/rails_5.1_stable.gemfile
   - gemfiles/rails_5.2_stable.gemfile
+  - gemfiles/rails_4.2_stable_mongoid.gemfile
+  - gemfiles/rails_5.0_stable_mongoid.gemfile
+  - gemfiles/rails_5.1_stable_mongoid.gemfile
+  - gemfiles/rails_5.2_stable_mongoid.gemfile
```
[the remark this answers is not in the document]
scicco: rails 6 added.

Appraisals
```diff
@@ -1,15 +1,35 @@
 appraise 'rails-4.2-stable' do
   gem 'rails', '~> 4.2.0'
+  gem 'sqlite3', '1.3.11'
```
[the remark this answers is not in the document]
scicco: I checked master branch. I've used

``````
group :active_record do
  gem "sqlite3", "~> 1.3.0"
end
``````
and
``````
group :mongoid do
  gem "mongoid", "~> 4.0"
end
``````

File: devise-security.gemspec [hunk 13]
```diff
@@ -44,5 +44,20 @@ Gem::Specification.new do |s|
   s.add_development_dependency 'pry'
   s.add_development_dependency 'rails_email_validator'
   s.add_development_dependency 'rubocop', '~> 0.59.2'
-  s.add_development_dependency 'sqlite3'
+  unless defined? DEVISE_ORM
```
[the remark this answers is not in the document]
scicco: ok I'll revert it to master version.

lib/devise-security/models/secure_validatable.rb
scicco: The original code (upstream/master) is:
``````
def has_uniqueness_validation_of_login?
  validators.any? do |validator|
    validator.kind_of?(ActiveRecord::Validations::UniquenessValidator) &&
    validator.attributes.include?(login_attribute)
  end
end
``````

my new proposal is:
``````
def has_uniqueness_validation_of_login?
  validators.any? do |validator|
    validator_orm_klass = DEVISE_ORM == :active_record ? 
    ActiveRecord::Validations::UniquenessValidator : ::Mongoid::Validatable::UniquenessValidator
    validator.kind_of?(validator_orm_klass) && validator.attributes.include?(login_attribute)
  end
end
``````
DEVISE_ORM can have only one value at time.

File: lib/devise-security/models/old_password.rb (hunk 3 above)
Comment (scicco) [reply lib/devise-security/models/old_password.rb]: I've splitted into two different files:

`lib/devise-security/models/active_record/old_password.rb`
`lib/devise-security/models/mongoid/old_password.rb`

and load the right one inside `devise-security.rb` with
`require "devise-security/models/#{DEVISE_ORM}/old_password" if defined? DEVISE_ORM`

**Comment by scicco** [comment]
I've cleaned the code and addressed your requested changes. I also pulled and merged last changes from upstream/master. The active_record tests should be all green, while the mongoid test are failing. When I test locally I still have two failing tests:

```
  1) Failure:
TestPasswordArchivable#test_updating_a_record_updates_the_time_the_password_was_changed_if_the_password_is_changed [/Users/g.scicolone/prj/devise-security/test/test_password_expirable.rb:67]:
Expected false to be truthy.

  2) Failure:
TestWithoutSecurityQuestion#test_When_security_question_is_not_enabled_it_is_not_inserted [/Users/g.scicolone/prj/devise-security/test/test_security_question_controller.rb:87]:
--- expected
+++ actual
@@ -1 +1 @@
-"You will receive an email with instructions for how to unlock your account in a few minutes."
+nil
50 runs, 135 assertions, 2 failures, 0 errors, 0 skips 
```
On travis I saw a stackoverflow error inside every mongoid [travis job](https://travis-ci.org/devise-security/devise-security/jobs/493806930), but i don't know how to replicate it. Even with wwtd runned from local I don't see that problem.

Any advice?

**Comment by olbrich** [comment]
@scicco please enable the '[Allow Edits from Maintainers](https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/)' option on your PR and I'll try to spend some time helping to fix some of the issues.

**Comment by olbrich** [comment]
@scicco ok, I've made some progress.  Still getting the stack too deep error, but there is another class of errors to address as well.  Those have to do with differences between the way active record and mongoid handle dirty record tracking.

**Comment by olbrich** [comment]
@scicco the 'stack level too deep' was caused by the password_archiveable callback.  It was adding an `OldPassword` to the `User` document when the password changed.  This re-triggered the `before_update` callback, which would attempt to re-archive the password.  There are some subtle differences in how callbacks work in active_record vs. mongoid.

**Comment by scicco** [comment]
Thank you @olbrich for your help. Now there are just two failing test ([test1](https://travis-ci.org/devise-security/devise-security/jobs/494689355), [test2](https://travis-ci.org/devise-security/devise-security/jobs/494689358)). They seems connected to the same issue related a missed unlock procedure. If i cleanup the mongoid models the tests seems to pass.

**Comment by olbrich** [comment]
> You may consider to use [database_cleaner](https://github.com/DatabaseCleaner/database_cleaner) instead of my ugly mongoid related hacks.

👍 Exactly my plan.

File: test/test_security_question_controller.rb [hunk 14]
```diff
@@ -58,6 +57,12 @@ class TestWithoutSecurityQuestion < ActionController::TestCase
   tests Devise::UnlocksController
 
   setup do
+    if DEVISE_ORM == :mongoid
```
scicco: this if is useless now that database_cleaner has been added

test/orm/20180515154723_create_mongoid_tables.rb
```diff
@@ -0,0 +1,93 @@
+# # frozen_string_literal: true
```
scicco: this file could be deleted

test/dummy/app/models/mongoid/all_fields.rb
```diff
@@ -0,0 +1,77 @@
+module AllFields
```
scicco: coveralls is decreasing because of this file. When I added this file I thought that having a single file loading all module fields could be useful. It can be excluded from coverage till it will be really used. @olbrich what do you think?

olbrich: @scicco if we don't use it, we should probably delete it.

olbrich:
@scicco take a look at the readme and add any additional information that might be useful to someone starting to use mongoid.  Also feel free to add yourself to the contributors list.

scicco:
@olbrich thank you again for your help. I was wondering if we can add a way to avoid to explicitly set DEVISE_ORM, or set the default value based whether mongoid gem is loaded or not, something like this
`
DEVISE_ORM = ENV.fetch('DEVISE_ORM', defined?(Mongoid) ? :mongoid : :active_record)
`

I saw that Devise uses this declaration inside his config file to select the proper ORM.

```
# ==> ORM configuration
  # Load and configure the ORM. Supports :active_record (default) and
  # :mongoid (bson_ext recommended) by default. Other ORMs may be
  # available as additional gems.
  require 'devise/orm/mongoid'
```

Any ideas?

olbrich:
> @olbrich thank you again for your help. I was wondering if we can add a way to avoid to explicitly set DEVISE_ORM, or set the default value based whether mongoid gem is loaded or not, something like this
> `DEVISE_ORM = ENV.fetch('DEVISE_ORM', defined?(Mongoid) ? :mongoid : :active_record)`
> 
> I saw that Devise uses this declaration inside config file to select the proper ORM.
> 
> ```
> # ==> ORM configuration
> # Load and configure the ORM. Supports :active_record (default) and
> # :mongoid (bson_ext recommended) by default. Other ORMs may be
> # available as additional gems.
> require 'devise/orm/mongoid'
> ```
> Any ideas?

I'm reluctant to implement this since Devise itself does not.  This may lead to some odd behavior if `devise` is loaded before `devise-security`.  

Right now I think it is possible to have both `active_record` and `mongoid` loaded and have `devise` use `active_record` as it's ORM.  Your proposal would break that use case.

It would be pretty easy to lock a rails app to set `DEVISE_ORM=:mongoid` in `application.rb` before `devise` and `devise-security` are required or bundler loads all of the other gems.

scicco:
I see. I added a Rails example inside the `README.md` file.

README.md
```diff
@@ -104,6 +104,37 @@ Devise.setup do |config|
...
+
+    brew install mongodb
+
+### Rails App setup example with Mongoid
+
+```ruby
+# inside config/application.rb
+require File.expand_path('../boot', __FILE__)
+#...
+DEVISE_ORM=:mongo_id
```
olbrich:
```suggestion
DEVISE_ORM=:mongoid
```

scicco: done

PR #79, opened 2019-02-14
Skip session limitable

kirichkov:
This updates the session_limitable strategy to allow bypassing the strategy altogether when this is desired, e.g. administrator impersonating a regular user.

After some back-and-forth I couldn't put the current implementation to use. I tried setting the warden.env and warden.request.env `devise.skip_session_limitable` key directly but to no avail to bring the desired results.

I've rewritten the relevant parts and retested, so one scope (administrator) impersonation another scope (user), does not have any effect on the primary user session and vice versa. If a user scope logs in (without impersonation), on a different browser, this properly interrupts the user session but not the admin's impersonating session. Unfortunately I don't think that's possible to test with rails' tests as it requires multiple concurrent, isolated one from one another requests.


oniofchaos:
There's an open PR https://github.com/devise-security/devise-security/pull/74 that appears to accomplish the same thing. Does this meet your needs?

kirichkov:
From looking at the patch it is different functionality with different goals. In #74 I think the skip flag will need to be persisted in the database, instead of persisting it in the session. This won't be useful in my case, as I want an admin to be able to impersonate users, but if the flag is persisted in the database, this flag will be applied to users too, not only to the impersonating admin's session.

#74 aims to allow certain users to have multiple sessions, my PR allows impersonation. We should figure out a way to merge those two features.

rusllonrails:
Any updates on this guys?

kirichkov:
We agreed that I'd rebase my work on top of #74 when it gets accepted. So far it doesn't look it's going anywhere :-(

oniofchaos:
@kirichkov #74 is merged!

kirichkov:
I just pushed some updated code, but since this can't be tested automatically using integration tests, let me do some manual tests, before merging.

lib/devise-security/hooks/session_limitable.rb
```diff
@@ -19,14 +24,13 @@
...
-  env   = warden.request.env
 
   if record.devise_modules.include?(:session_limitable) &&
      warden.authenticated?(scope) &&
      options[:store] != false
     if record.unique_session_id != warden.session(scope)['unique_session_id'] &&
-       !env['devise.skip_session_limitable'] &&
-       !record.skip_session_limitable?
+       !record.skip_session_limitable? && 
+       !warden.session(scope)['devise.skip_session_limitable']
```
n00dle, line 33: Prior to this change, we were able to modify the request by prepending `Devise::SessionController`'s `create` action, to trigger the  `skip_session_limitable` behavior:
```ruby
request.env['devise.skip_session_limitable'] = true
```

The (effective) change on this line no longer allows us this control:
```diff
-       !warden.request.env['devise.skip_session_limitable']
+       !warden.session(scope)['devise.skip_session_limitable']
```

Is there still a way to trigger the `skip_session_limitable` functionality in a similar manner?

dillonwelch: Hi @n00dle, I only just saw this comment. If this is still a problem for you, can you open up a new issue and we can discuss further there?

victorfgs: Just a +1: I had the same problem here. With this change, preprended actions on my admin session controller do not work anymore to skip the limitable hook.

It's not really a problem since I can use the functionality as documented in the readme but I thought it would be interesting to document here that some devise hooks (like `trackable`) have that same functionality to skip the hook and implemented this by using the `warden.request.env` params + controller `prepended_before_action`. It seems weird to me that this gem uses the same hook definition but by using other mechanisms. In my case, I was using the `bypass_sign_in` method and since it does not accept keyword args I did not have a way to set the skip variable anymore.
If warden callbacks were a problem to me, it would result in a dead-end

dillonwelch: https://github.com/devise-security/devise-security/issues/421 please put any further discussion in here please! <3

Don't have the time to take a look into it right now unfortunately.

PR #80, opened 2019-02-19
pin rmagick & adjust travis builds

olbrich:
rmagick released a new version recently that fails to install on travis.  This branch pins to a lower version until that can be resolved.

I also cleaned up the travis configs a bit so that it does not perform unnecessary builds.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,17 +9,23 @@ rvm:
   - 2.6
   - ruby-head
 matrix:
-  allow_failures:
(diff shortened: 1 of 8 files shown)


oniofchaos:
Link to the failed build?

olbrich:
@oniofchaos https://travis-ci.org/devise-security/devise-security/builds/495263212.  FYI, A while back I setup travis to run a build on master about once a month to help detect when new dependency releases break things.

PR #84, opened 2019-02-22
Add installation of imagemagick to CI so that rmagick can be unlocked

oniofchaos:
Follow up to https://github.com/devise-security/devise-security/pull/80

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: ruby
 dist: xenial
-# before_install:
-#   # pin to bundler < 2 because Rails 4.2 is incompatible
-#   - gem install bundler -v '1.17.3'
+before_install:
+  # install bundler < 2 because Rails 4.2 is incompatible with bundler >= 2
+  - gem install bundler -v '1.17.3'
 # install: bundle _1.17.3_ install --jobs=2 --retry=2 --path=${BUNDLE_PATH:-vendor/bundle}
 cache: bundler
 script: bundle exec rake
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
 language: ruby
 dist: xenial
-before_install:
-  # pin to bundler < 2 because Rails 4.2 is incompatible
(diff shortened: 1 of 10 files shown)


.travis.yml
```diff
@@ -1,5 +1,15 @@
 language: ruby
-before_install: gem install bundler -v '1.17.3'
+before_install:
+  - gem install bundler -v '1.17.3'
+  - sudo apt-get install build-essential pkg-config
```
olbrich: @oniofchaos A couple of things....
First, Thanks for looking into this!
Second, this adds several minutes to the build while it compiles imagemagick.  Perhaps we can find a way to cache the installation.  This stack overflow answer has some hints about how we can do that... https://stackoverflow.com/questions/41138404/how-to-install-newer-imagemagick-with-webp-support-in-travis-ci-container.
I'll also point out that rmagic has a script in the repo that does the dependency install they use during their circleci build.  It might be possible to leverage that script.   I tried experimenting with doing a `gem unpack rmagic` and then executing the script, but ran into a couple of issues.

oniofchaos: That brings up an interesting point. I looked through the initial build setup from Travis CI and it appears that it has ImageMagick already, so this might be just an issue of it not being within the global path and/or the version it has is not the version that rmagick 3 wants. 

screen shot 2019-02-24 at 5 07 45 pm

olbrich: I think their base image does not have the version that rmagick wants.  One way to handle this might be to file a request with Travis to update it.  I'm sure this is breaking builds for lots of people besides us.

.travis.yml
```diff
@@ -1,5 +1,15 @@
 language: ruby
-before_install: gem install bundler -v '1.17.3'
```
olbrich: My experiments suggest that if you add
```
dist: xenial
```
and then drop all the rmagick specific install steps, that this will just work.

oniofchaos:
Nice!!! Thanks for cleaning this up.

PR #85, opened 2019-02-22
Add command to cache bundler gems

oniofchaos:


.travis.yml
```diff
@@ -30,7 +26,6 @@ matrix:
     - rvm: ruby-head
     - gemfile: gemfiles/rails_6.0_beta.gemfile
 gemfile:
-  - gemfiles/rails_4.2_stable.gemfile
```
oniofchaos: Removed this since Rails 4.2 doesn't support Bundler 2 and Travis CI installs bundler 2 by default. Rails 4.2 only receives severe security updates at this point, and will be officially EOL once Rails 6 is out. 

Thoughts?

olbrich: I'm fine dropping Rails 4.2 *after* it's officially EOL'd.  That probably won't be too long now considering that Rails 6.0 is in beta.

gemfiles/rails_4.2_stable.gemfile
```diff
@@ -2,7 +2,7 @@
 
 source "https://rubygems.org"
 
-gem "bundler", "< 2"
```
olbrich: for consistency, you should delete this file

oniofchaos: Not gonna delete for now because I'm leaving this in until Rails 4.2 is officially EOL

.travis.yml
```diff
@@ -1,6 +1,7 @@
 language: ruby
-before_install: gem install bundler -v '1.17.3'
-install: bundle _1.17.3_ install --jobs=3 --retry=3
+before_install: gem install bundler -v '1.17.2'
+install: bundle _1.17.2_ install --jobs=2 --retry=2 --path=${BUNDLE_PATH:-vendor/bundle}
```
olbrich: Why did we go from bundler 1.17.3 to 1.17.2?

oniofchaos: This was required to get the Ruby 2.6 builds to work - I think because Ruby comes with it bundled in now. It was throwing a mismatched versions error.

olbrich: weird, the `before_install` script should have forced installation of the newest version and then the `install` script should have used that specific version.  In any event, it's not a major difference and I don't think it impacts our usage.

oniofchaos: Agreed. In any case, once we can drop Rails 4.2 support we can get rid of this line entirely and it will automatically use Bundler 2.0.

PR #87, opened 2019-03-04
Don't check in appraisal-generated gemfiles.

olbrich:
Generate them every time the build runs.

This way if dependabot updates a dependency we will actually do a build that tests the new dependency.


olbrich:
This doesn't seem work in a reliable way on travis.  Closing.

oniofchaos:
Neat idea though!

PR #71, opened 2019-01-15
 Support logging of sessions with flexible session limits

itsmechlark:
Allows having a history of login sessions with flexible active session limits.

PR #98, opened 2019-04-13
Cleanup session limitable

olbrich:
This is a refactor / cleanup of the `session_limitable` feature and is a pre-requisite for #71.

This also fixes a bug where an invalid User record would have been saved with invalid data and where one would have been created invalid.


test/test_session_limitable.rb
```diff
@@ -0,0 +1,31 @@
+require 'test_helper'
+
+class TestSessionLimitable < ActiveSupport::TestCase
```
oniofchaos: Based on your good point about missing test coverage - what do you think about adding tests for this module first in a separate PR and then refactoring?

olbrich: I'm trying to keep the refactors to a minimum, but I can pull them out entirely if you think it helps keep things cleaner.

oniofchaos: What you have here is fine to me. If this was going to be a long `WIP` branch that's when my suggestion comes in.

olbrich: 👍 Once I finish up another subset of tests it should be good to go.

lib/devise-security/hooks/session_limitable.rb
```diff
@@ -12,15 +11,17 @@
...
+# another browser was opened for the record or not, based on a unique session
+# identifier. If so, the old account is logged out and redirected to the sign in
+# page on the next request.
 Warden::Manager.after_set_user only: :fetch do |record, warden, options|
   scope = options[:scope]
   env   = warden.request.env
 
   if record.respond_to?(:unique_session_id) && warden.authenticated?(scope) && options[:store] != false
     if record.unique_session_id != warden.session(scope)['unique_session_id'] && !env['devise.skip_session_limitable']
+      Rails.logger.warn { "[devise-security][session_limitable] session id mismatch: expected=#{record.unique_session_id.inspect} actual=#{warden.session(scope)['unique_session_id'].inspect}" }
```
oniofchaos: Can you multi-line this so we don't have a giant line?

lib/devise-security/models/session_limitable.rb
```diff
@@ -12,10 +12,16 @@ module Models
     module SessionLimitable
       extend ActiveSupport::Concern
 
+      # Update the unique_session_id on the model.  This will be checked in
+      # the Warden after_set_user hook in {file:devise-security/hooks/session_limitable}
+      # @param unique_session_id [String]
+      # @return [void]
```
oniofchaos: `null` over `void` in ruby?

olbrich: @return [void] is yard's way of indicating that the return value is not significant.

gemfiles/rails_6.0_beta.gemfile
```diff
@@ -2,7 +2,7 @@
 
 source "https://rubygems.org"
 
-gem "rails", "~> 6.0.0.beta1"
```
natebird: Should this be changed to RC1 now that it is released?

test/test_session_limitable.rb
natebird: Have the other tests been wrapped up?

Issue #103, opened 2019-05-14
uninitialized constant Devise::Models::Compatibility::ActiveRecord::...


oehlschl:
Just started experiencing a regression after updating from 0.13->0.14.1 and traced it back to the commit introducing Mongo support: #76 

It looks like the new `Devise::Models::Compatibility::ActiveRecord` module now conflicts with `::ActiveRecord`. Specifically, accessing any descendent of `ActiveRecord::` within a model that includes devise-security now requires absolute pathing.

For example,

```ruby
# in user.rb, after including devise-security

ActiveRecord::Base.transaction do
   # ...
end

# => NameError: uninitialized constant Devise::Models::Compatibility::ActiveRecord::Base

raise ActiveRecord::RecordInvalid

# => NameError: uninitialized constant Devise::Models::Compatibility::ActiveRecord::RecordInvalid
```

The errors can be circumvented by accessing the root-level ActiveRecord explicitly, like `raise ::ActiveRecord::RecordInvalid`, but that's surprising and also not idiomatic.

I'll open a PR that replicates the error momentarily.

PR #104, opened 2019-05-14
Fix namespace resolution problems for compatibility modules

oehlschl:
Some quick specs to demonstrate the issue reported in #103


oehlschl:
Thanks for the quick changes, @olbrich. For what it's worth, they look good to me!

PR #111, opened 2019-07-03
Allow API responses in PasswordExpiredController

Slike9:
For API sometimes it's desirable to receive 2XX responses instead of
3XX. The same technique is used in `devise` gem.

More about the problem addressed: https://github.com/devise-security/devise-security/pull/111#issuecomment-508189297


olbrich:
@Slike9 Thanks for your contribution, please add a link to the relevant part of the devise codebase, add some tests, and add some comments describing why this is useful.  We also need to consider how this might break code that depends on the existing behavior.

Slike9:
@olbrich 
> please add a link to the relevant part of the devise codebase

Similar code `respond_with({}, location: ...)` is used in several places in `devise` codebase:
* https://github.com/plataformatec/devise/blob/master/app/controllers/devise/confirmations_controller.rb#L15
* https://github.com/plataformatec/devise/blob/master/app/controllers/devise/passwords_controller.rb#L19
* https://github.com/plataformatec/devise/blob/master/app/controllers/devise/unlocks_controller.rb#L17

> add some comments describing why this is useful

Method `respond_with` from `responders` gem is applied to be able to respond to different requested formats (html, json, etc.).
For formats associated with browsing, like `:html`, `:iphone`, etc, the fix won't change the behavior, the action will respond with a redirect as before.
The behavior will change for formats associated with APIs, such as `:xml` and `:json`, the action will respond with `204 No Content` instead of redirect.

The problem we are facing is the following. In our SPA-like application, we access the controller's actions via AJAX, requesting `json` format. To update password, the client side requests `PUT /users/password_expired.json` via AJAX. The browser gets `302 Location: /` and follows the redirect to `/` using PUT method (it preserves the verb) and gets 404. In the client side nothing happens, as it doesn't see 302 response, just 404. More about this problem:
* https://stackoverflow.com/questions/228225/prevent-redirection-of-xmlhttprequest
* https://github.com/axios/axios/issues/932#issuecomment-307390761

So, we need API friendly responses to AJAX requests, not 302.

> add some tests

Do we need to test behavior of `respond_with` of `responders` gem? The behavior for html format is already tested, isn't it enough?

test/controllers/test_password_expired_controller.rb
```diff
@@ -6,13 +6,14 @@ class Devise::PasswordExpiredControllerTest < ActionController::TestCase
...
   setup do
+    @controller.class.respond_to :json, :xml
     @request.env["devise.mapping"] = Devise.mappings[:user]
     @user = User.create!(
       username: 'hello',
       email: 'hello@path.travel',
       password: 'Password4',
       password_changed_at: 4.months.ago,
-      confirmed_at: 5.months.ago
+      confirmed_at: 5.months.ago,
```
oniofchaos: Remove unnecessary comma

Appraisals
```diff
@@ -8,27 +8,39 @@ end
 
 appraise 'rails-5.0-stable' do
   gem 'rails', '~> 5.0.0'
+  group :test do
+    gem 'rails-controller-testing'
```
oniofchaos: What was this gem added for? This is an outdated method of testing and generally meant to support legacy apps, not for new code.

olbrich: We have some tests that need to run against controllers in Rails 4.2.  This lets them work against rails 5+ as well.

test/controllers/test_password_expired_controller.rb
olbrich:
```suggestion
      confirmed_at: 5.months.ago
```

test/controllers/test_password_expired_controller.rb
```diff
@@ -23,24 +24,116 @@ class Devise::PasswordExpiredControllerTest < ActionController::TestCase
...
-        }
-      }
+  test 'update password with default format' do
+    if Rails.gem_version < Gem::Version.new('5.0')
+      put :update,
+          {
+            user: {
+              current_password: 'Password4',
+              password: 'Password5',
+              password_confirmation: 'Password5',
```
olbrich:
```suggestion
              password_confirmation: 'Password5'
```

test/controllers/test_password_expired_controller.rb
```diff
@@ -23,24 +24,116 @@ class Devise::PasswordExpiredControllerTest < ActionController::TestCase
...
-          password: 'Password5',
-          password_confirmation: 'Password5'
-        }
-      }
+      put :update,
+          params: {
+            user: {
+              current_password: 'Password4',
+              password: 'Password5',
+              password_confirmation: 'Password5',
```
olbrich:
```suggestion
              password_confirmation: 'Password5'
```

test/controllers/test_password_expired_controller.rb
```diff
@@ -23,24 +24,116 @@ class Devise::PasswordExpiredControllerTest < ActionController::TestCase
...
+  end
+
+  test 'password confirmation does not match' do
+    if Rails.gem_version < Gem::Version.new('5.0')
+      put :update,
+          {
+            user: {
+              current_password: 'Password4',
+              password: 'Password5',
+              password_confirmation: 'Password6',
```
olbrich:
```suggestion
              password_confirmation: 'Password6'
```

test/controllers/test_password_expired_controller.rb
```diff
@@ -23,24 +24,116 @@ class Devise::PasswordExpiredControllerTest < ActionController::TestCase
...
+            }
+          }
+    else
+      put :update,
+          params: {
+            user: {
+              current_password: 'Password4',
+              password: 'Password5',
+              password_confirmation: 'Password6',
```
olbrich:
```suggestion
              password_confirmation: 'Password6'
```

test/controllers/test_password_expired_controller.rb
```diff
@@ -23,24 +24,116 @@ class Devise::PasswordExpiredControllerTest < ActionController::TestCase
...
+    if Rails.gem_version < Gem::Version.new('5.0')
+      # The responders gem that is compatible with Rails 4.2
+      # does not return a 204 No Content for common data formats
+      # This is the previously existing behavior so it is allowed
+      put :update,
+          {
+            user: {
+              current_password: 'Password4',
+              password: 'Password5',
+              password_confirmation: 'Password5',
```
olbrich:
```suggestion
              password_confirmation: 'Password5'
```

test/controllers/test_password_expired_controller.rb
```diff
@@ -23,24 +24,116 @@ class Devise::PasswordExpiredControllerTest < ActionController::TestCase
...
+      assert_redirected_to root_path
+      assert_equal response.content_type, 'text/html'
+    else
+      put :update,
+          format: :json,
+          params: {
+            user: {
+              current_password: 'Password4',
+              password: 'Password5',
+              password_confirmation: 'Password5',
```
olbrich:
```suggestion
              password_confirmation: 'Password5'
```

test/controllers/test_password_expired_controller.rb
```diff
@@ -23,24 +24,116 @@ class Devise::PasswordExpiredControllerTest < ActionController::TestCase
...
+    if Rails.gem_version < Gem::Version.new('5.0')
+      # The responders gem that is compatible with Rails 4.2
+      # does not return a 204 No Content for common data formats
+      # This is the previously existing behavior so it is allowed
+      put :update,
+          {
+            user: {
+              current_password: 'Password4',
+              password: 'Password5',
+              password_confirmation: 'Password5',
```
olbrich:
```suggestion
              password_confirmation: 'Password5'
```

test/controllers/test_password_expired_controller.rb
```diff
@@ -23,24 +24,116 @@ class Devise::PasswordExpiredControllerTest < ActionController::TestCase
...
+      assert_redirected_to root_path
+      assert_equal response.content_type, 'text/html'
+    else
+      put :update,
+          format: :xml,
+          params: {
+            user: {
+              current_password: 'Password4',
+              password: 'Password5',
+              password_confirmation: 'Password5',
```
olbrich:
```suggestion
              password_confirmation: 'Password5'
```

billgloff:
Would love to see this get merged. Let me know if you need any help.

billgloff:
In `lib/devise-security/controllers/helpers.rb` there are two checks (`handle_password_change` & `handle_paranoid_verification`) for checking if the format is html, can we remove that and handle both cases (html and json) and just return the proper error message if its json format so the frontend can handle it properly?

olbrich:
@billgloff can you elaborate a bit on how you would like the `helpers` to respond?  Some failing tests would be extremely helpful.  As an alternative, you could open a separate ticket / PR for your modifications and then I'll go ahead and merge this branch.

oniofchaos:
@olbrich how do you feel about merging this and then we can handle the request in a separate PR as desired?

PR #115, opened 2019-07-11
View translations

D-system:
Only moving the harded translation from the views to the translation file.


oniofchaos:
@ D-system LGTM and thanks for the contribution. Can you merge in latest master so we can merge in the PR?

D-system:
@oniofchaos Updated to the latest master

PR #117, opened 2019-07-17
Fix typo in ja locale

hiroki-sawano:
Hi, I found a typo in `invalid_security_question` and fixed it.

### DIFF
--- a/config/locales/ja.yml
+++ b/config/locales/ja.yml
@@ -18,7 +18,7 @@ ja:
           other: は最低%{count}つの大文字英字を含む必要があります。
   devise:
     invalid_captcha: 'キャプチャ入力が不正です。'
-    invalid_security_question: 'セキュリティ質問に対すつ回答が不正です。'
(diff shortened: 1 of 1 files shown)


config/locales/ja.yml
```diff
@@ -18,7 +18,7 @@ ja:
           other: は最低%{count}つの大文字英字を含む必要があります。
   devise:
     invalid_captcha: 'キャプチャ入力が不正です。'
-    invalid_security_question: 'セキュリティ質問に対すつ回答が不正です。'
+    invalid_security_question: 'セキュリティ質問に対する回答が不正です。'
```
oniofchaos: Confirming that the typo fix is `つ` => `る`? Can you explain what the difference is?

hiroki-sawano: Yes, `対すつ` should be `対する` that means `to` as in "The answer `to` the security question was invalid". `対すつ` just doesn't make sense because of the typo.

D-system: I'm not native Japanese but I confirm, it's `る` 100%

PR #119, opened 2019-07-23
Compatibility with Rails 4 - ApplicationRecord vs. ActiveRecord::Base

agronv:
This PR https://github.com/devise-security/devise-security/pull/76 intended to add Mongoid support had this commit https://github.com/devise-security/devise-security/commit/596696d2df4eada853ec72c32018449252aacad0#diff-9fd2d1ec60d420f04b7daa81b3fa653d that was Rails 5 specific.

### DIFF
--- a/lib/devise-security/models/active_record/old_password.rb
+++ b/lib/devise-security/models/active_record/old_password.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
-class OldPassword < ApplicationRecord
+class OldPassword < ActiveRecord::Base
   belongs_to :password_archivable, polymorphic: true
 end


lib/devise-security/models/active_record/old_password.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
-class OldPassword < ApplicationRecord
```
oniofchaos: This should have a check for Rails 5 and use `ApplicationRecord` if it's supported.

oniofchaos: Also should have a test

olbrich:
There's an easy work around for this...
Just declare `class ApplicationRecord < ActiveRecord::Base; end`.  There isn't anything Rails 5 specific about ApplicationRecord AFAIK.

oniofchaos:
Closing this since Rails 4.2 is EOL'd. https://github.com/rails/rails/commit/1c3c4f053f5b00dce5ed53570b2938977b830289

PR #124, opened 2019-08-18
Update travis build to support Rails 6

olbrich:
* added a maintenance policy
* indicated intention to drop support for Rails 4.2 when Ruby 2.4 is EOL


README.md
```diff
@@ -322,6 +322,12 @@ To locally simulate what travis-ci will run when you push code use:
     $ gem install bundler -v '1.17.3'
     $ BUNDLER_VERSION=1.17.3 wwtd
 
+## Maintenance Policy
+
+We are committed to maintain support for `devise-security` for all normal or security maintenance versions of the Ruby language [as listed here](https://www.ruby-lang.org/en/downloads/branches/), and for the Ruby on Rails framework [as per their maintenance policy](https://rubyonrails.org/maintenance/).
+
+Following the release of Rails 6, support for Rails 4.2 will be dropped when Ruby 2.4 reaches EOL on 2020-03-31.
```
oniofchaos: I think we can drop Rails 4.2 support now since it's EOL'd by Rails, regardless of Ruby version.

https://github.com/rails/rails/commit/1c3c4f053f5b00dce5ed53570b2938977b830289

olbrich: We could, I am suggesting we drop support for Rails 4.2 when Ruby 2.4 ships because we haven't clearly indicated our policy and I wanted to give users a reasonable amount of time to adjust.  The Ruby 2.4 EOL is a arbitrary date and is not a technical one.

oniofchaos: Removing Rails 4.2 support can go along with bumping to `0.15.0`, or maybe `0.16.0`? Given that it's EOL'd, I think we should remove it sooner rather than later. Tying it to a version bump is sufficient I think in terms of notice - basically saying Rails 4.2 is supported up until X version.

PR #132, opened 2019-09-16
handle password confirmation mismatched

sanjsharma:
At the moment, when the password expires and user is taken to new password page.  If the user types in two different passwords in `New Password` and `Confirm Password` the code does check if these two match or not. 

This pull request solves this issue. Cheers !

### DIFF
--- a/config/locales/de.yml
+++ b/config/locales/de.yml
@@ -1,6 +1,7 @@
 de:
   errors:
     messages:
+      password_confirmation_mismatched: 'nicht übereinstimmend.'
       taken_in_past: 'wurde bereits in der Vergangenheit verwendet.'
       equal_to_current_password: 'darf nicht dem aktuellen Passwort entsprechen.'
       password_complexity:
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1,6 +1,7 @@
 en:
   errors:
     messages:
+      password_confirmation_mismatched: 'mismatched.'
       taken_in_past: 'was used previously.'
       equal_to_current_password: 'must be different than the current password.'
       password_complexity:
--- a/config/locales/es.yml
+++ b/config/locales/es.yml
@@ -1,6 +1,7 @@
 es:
   errors:
     messages:
+      password_confirmation_mismatched: 'no coinciden.'
       taken_in_past: 'la contraseña fue usada previamente, favor elegir otra.'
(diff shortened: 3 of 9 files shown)


tmr08c:
When trying to add a test for this, I saw what looked like duplicate errors around password confirmations:

```html
<div id="error_explanation">
  <h2>
    4 errors prohibited this user from being saved:
  </h2>
  <ul>
      <li>Password confirmation doesn&#39;t match Password</li>
      <li>Password confirmation mismatched.</li>
      <li>Current password is invalid</li>
      <li>Password is invalid</li>
  </ul>
</div>
```

In debugging this, I saw that `User#valid?` would result in an error about the password confirmation before adding the new custom error message:

```ruby
[1] pry(#<User>)> self.valid?
=> false
[2] pry(#<User>)> errors
=> #<ActiveModel::Errors:0x00007fbe7e5fc5c8
 @base=
  #<User id: 1, username: "user1", facebook_token: nil, unique_session_id: nil, email: "hello1@path.travel", password_changed_at: "2020-02-18 00:07:59", created_at: "2020-06-18 00:08:00", updated_at: "202
0-06-18 00:08:00", paranoid_verification_code: nil, paranoid_verified_at: nil, paranoid_verification_attempt: 0, security_question_id: nil, security_question_answer: nil, expired_at: nil, last_activity_at
: "2020-06-18 00:08:00">,
 @details={:password_confirmation=>[{:error=>:confirmation, :attribute=>"Password"}]},
 @messages={:password_confirmation=>["doesn't match Password"]}>
```

After doing some digging, it looks like this implemented in `Devise::Validatable`:

https://github.com/heartcombo/devise/blob/4f60544396824c516bdb5c165f9631d727b9560d/lib/devise/models/validatable.rb#L41

They are leveraging `validate_confirmation_of` which is a Rails helper defined here:

https://github.com/rails/rails/blob/661da266b94909574426fd1121ef13b800e01b9a/activemodel/lib/active_model/validations/confirmation.rb#L75


@sanjsharma are you including `Devise::Validatable` in your `User` model (or whichever model has the password)?

You can see how we are including it in our `User` model in the app used for testing:

https://github.com/devise-security/devise-security/blob/c467f0c9353213c36fd6732855caf672a8102b19/test/dummy/app/models/user.rb#L5-L21

PR #173, opened 2020-02-25
Fix attribute_changed? deprecation warning

taylor-au:
Fix attribute_changed? deprecation warning


oniofchaos:
@taylor-au why did you close?

taylor-au:
@oniofchaos I was thinking that it would be a trivial code change to remove the deprecation warning, but it looks like it may require a Ruby upgrade and I don't have the time to invest in fixing this at the moment.

oniofchaos:
No problem!

oniofchaos:
BTW a thought - you can do something like
```ruby
if (defined?(:will_save_change_to_encrypted_password?) && will_save_change_to_encrypted_password?) ...
end
```
and that should handle older versions ruby

PR #195, opened 2020-04-28
Simpler valid email for PasswordExpiredControllerTest

jvhaperen:
After initial test suite run before doing another PR - so without any changes - I run into the issue of a failing test in PasswordExpiredControllerTest because of the used email for user. Since the aim of this specific test is not the email-validator, I thought it would be a good idea to use a simpler email address there, so that future contributors wont run into the same issue.

I also changed `create!` to `create` so if the user validation fails, this test will fail on the assert instead of getting the raise from the invalid record on.

### DIFF
--- a/test/controllers/test_password_expired_controller.rb
+++ b/test/controllers/test_password_expired_controller.rb
@@ -8,9 +8,9 @@ class Devise::PasswordExpiredControllerTest < ActionController::TestCase
   setup do
     @controller.class.respond_to :json, :xml
     @request.env["devise.mapping"] = Devise.mappings[:user]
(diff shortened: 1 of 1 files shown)


olbrich:
@jvhaperen I haven't been able to reproduce the test failure that prompted this PR, can you provide some additional information about it?

jvhaperen:
Can't reproduce anymore as well :S Maybe the email validator gem (rails_email_validator) failed on its mx or idn check that it does by default, at the time I ran the tests (multiple times).

olbrich:
closing this since we can't reproduce it

PR #196, opened 2020-04-28
Add Dutch (NL) locale

jvhaperen:
Add Dutch (NL) locale.
Instead of only having this in my local app I thought it would be nice to add this to Devise-security itself.


olbrich:
@jvhaperen thanks for your contribution!

PR #201, opened 2020-05-16
password_expirable helper bugfix

### LINKED ISSUE #186: Occasionally getting falsy "Password expired" msg
Until now we were using `devise_security_extensions` 0.10.0
which had this issue and had been fixed with a fork. 
see commit: https://github.com/OwnBackup/devise_security_extension/commit/3557f8fb01b0d2fee56c47fa6b0a1a32b74e876f#diff-319d340c1a338ea901c51a01e3cc5004

Now we've upgraded to latest `devise_security` and the same code causing us this issue is still there.
In order not to fork again, I've created a patchy rails initializer in our system for now:
```ruby
module PassExpFix                                                                                  
  def []= (k, v)                                                                                   
    Rails.logger.debug "##### warden.sesseion(:user)[#{k.inspect}] = #{v.inspect} ##### has been called"
    if k == :password_expired                                                                      
      k = 'password_expired'
    end 
    super(k, v)                                                                                    
  end
end

module PassExpWardenPatch                                                                          
  def session(scope)
    if scope == :user                                                                              
      instance = super(scope)                                                                      
      instance && instance.extend(PassExpFix)
    else
      super                                                                                        
    end
  end
end

class Warden::Proxy                                                                                
  prepend PassExpWardenPatch
end
```

olbrich:
Fixes #186

Co-authored-by: Troy Rosenberg <tmr08c@gmail.com>


lib/devise-security/controllers/helpers.rb
```diff
@@ -40,71 +40,80 @@ def valid_security_question_answer?(resource, answer)
...
           end
         end
+      end
 
-        # lookup if extra (paranoid) code verification is needed
-        def handle_paranoid_verification
-          return if warden.nil?
+      # lookup if extra (paranoid) code verification is needed
+      def handle_paranoid_verification
+        return if warden.nil?
```
dan-jensen, line 76: Hi @olbrich.  I'm trying to understand what this PR does to fix the bug, but I'm finding it difficult to identify what functionally changed.  Very respectfully, could I suggest making the formatting changes in another PR, so the bugfix here is as clear as possible?  P.S. I really appreciate the work you're doing the fix the bug, as I'm affected by it - thank you!

olbrich: @dan-jensen Yeah, I'll clean this up a bit when I work on it again on Wednesday.
BTW, we have been struggling with coming up with scenarios to reproduce the original problem, so if you have more details that would be appreciated.

You are also welcome to join us for a LiveShare coding session, I'll add a link here on Wednesday (2020-05-20) night at about 8pm Eastern.

olbrich:
I'm starting an open source collaboration session on this, join us if you are interested.

https://prod.liveshare.vsengsaas.visualstudio.com/join?87532C26E73FFEDBF25DDA7881512A3D39B9

To join the video meeting, click this link: https://meet.google.com/kmt-ffpn-gvf

test/controllers/test_widgets_controller.rb
```diff
@@ -0,0 +1,26 @@
...
+      password_changed_at: 4.months.ago,
+      confirmed_at: 5.months.ago
+    )
+    assert @user.valid?
+    assert @user.need_change_password?
+
+    sign_in(@user)
+  end
+
+  test 'should render show' do
```
tmr08c, line 22: Was this test added to help debug, or was there something it helped validate? 

I don't recall why the user wouldn't be redirected to `user_password_expired_path` like in the other test.  I tried poking around and adding an `assert_redirected_to`, and confirmed it was **not** redirecting.

test/controllers/test_password_expired_controller.rb
```diff
@@ -16,6 +16,8 @@ class Devise::PasswordExpiredControllerTest < ActionController::TestCase
       confirmed_at: 5.months.ago
     )
     assert @user.valid?
+    assert @user.need_change_password?
```
tmr08c, line 19: I think this was added while hunting down the problem. Seems reasonable to keep.

olbrich: This was mostly just assuring that our assumptions were being met.

lib/devise-security/hooks/password_expirable.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+# @note This happens after
+#   {DeviseSecurity::Controller::Helpers#handle_password_change}
```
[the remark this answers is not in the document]
olbrich: @natebird figuring out what was going on was 90% of this bugfix

test/controllers/test_widgets_controller.rb
olbrich: This was added to try and track down the problem, but this may now be superseded by other tests. It doesn't redirect because `WidgetsController` does not inherit from `Devise::PasswordExpiredController`

PR #216, opened 2020-06-27
Showing warning message for multiple for gem rails-controller-testing

anoobbava:
Hi Team,

when I cloned the gem and tried for bundle install, it shows the below error,

`Your Gemfile lists the gem rails-controller-testing (>= 0) more than once.
You should probably keep only one of them.
Remove any duplicate entries and specify the gem only once (per group).
While it's not a problem now, it could cause errors if you change the version of one of them later.`
Screen Shot 2020-06-27 at 11 52 49 AM

Please check if it is valid or please do discard.


devise-security.gemspec
```diff
@@ -18,7 +18,7 @@ Gem::Specification.new do |s|
     'Alexander Dreher',
     'Nate Bird',
     'Dillon Welch',
-    'Kevin Olbrich'
+    'Kevin Olbrich',
```
tmr08c, line 21:
```suggestion
    'Kevin Olbrich'
```

It looks like this may have been autocorrected. It's probably fine to leave it in, but I figured I'd suggest keeping it as it was 🤷

anoobbava: @tmr08c Sure, Will add a new commit for that, Thanks.

anoobbava: @tmr08c  Added the commit now, please check.

tmr08c:
Thank you for updating the PR and thank you again for the contribution, @anoobbava!

PR #219, opened 2020-06-29
Remove rails requirement

vimutter:
Hi, 

requiring whole rails is not optimal since it limits applications from dropping some of rails optional gems from Gemfile. 
This gem only uses Engine and `active_support` both of which are required by devise's dependency - `railties`.


olbrich:
@vimutter can you please describe which rails components you need to drop?

vimutter:
> @vimutter can you please describe which rails components you need to drop?

Our use case was that we were not using `actioncable`, `actionmailbox`, `actiontext`, `activejob` and `activestorage`.

Gemfile.lock
```diff
@@ -267,7 +266,7 @@ DEPENDENCIES
   pry-byebug
   pry-rescue
   rails (~> 5.0.0)
```
oniofchaos, line 268: Was this file regenerated properly? This shouldn't be listed as a top level dependency if it's not in the `gemspec` anymore.

olbrich: Yes, the Gemfile specifies a Rails version that we use for local development, it's not part of the gemspec

olbrich (approved):

Gemfile.lock
olbrich: actually, I have updated the appraisals to use railties instead of the full rails, so this should go away now.

oniofchaos (approved):

Issue #232, opened 2020-10-04
🎃 Hacktoberfest 2020 


tmr08c:
Happy [Hacktoberfest](https://hacktoberfest.digitalocean.com/) everyone! 

# What to work on 

While we have a [few Issues](https://github.com/devise-security/devise-security/issues?q=is%3Aissue+is%3Aopen+label%3AHacktoberfest) that are leveraging the `Hacktoberfest` label, these are not the only way you can help contribute. 

You can also consider:

* Adding or improving our [translations](https://github.com/devise-security/devise-security/tree/master/config/locales) (see #196 and #223 for examples). 
* Addressing [Code Climate issues](https://codeclimate.com/github/devise-security/devise-security/issues).

# How to contribute

Please take a look at our [section on contributing](https://github.com/devise-security/devise-security#contributing-to-devise-security) in the README for help getting started. 

# Getting support 

Feel free to leverage this Issue as a thread for sharing ideas or reaching out for support.

dingyifei:
Hi, can I work on the `zh-CN` translation?

tmr08c:
@dingyifei, yeah, that'd be great!

dingyifei:
#235 Done!

oniofchaos:
Looks like @dingyifei is correct.

https://hacktoberfest.digitalocean.com/hacktoberfest-update

```
PRs count if:
Submitted in a repo with the hacktoberfest topic AND
during the month of October AND (
  The PR is merged OR
  The PR is labelled as hacktoberfest-accepted by a maintainer OR
  The PR has been approved
)
```

mhi20:
Hi,
Can I add a `fa` locale?

tmr08c:
Thanks for adding the topic, @oniofchaos!

@mhi20, definitely!

tmr08c:
I hope everyone had a successful Hacktoberfest. A big thank you to everyone that opened up PRs in our project this year!

PR #235, opened 2020-10-04
Add Simplified Chinese(zh_CN) locale

dingyifei:
#232


config/locales/zh_CN.yml
```diff
@@ -0,0 +1,40 @@
+en:
```
olbrich, line 1: I think this is required to prevent overwriting all the english terms. I think this is what is causing the build to fail.

```suggestion
zh_CN:
```

olbrich (changes requested):
Thank you for your contribution, I think there's only one thing we need to fix before we can accept this.

dingyifei:
Thank you, could you re-run the Travis-ci?

olbrich (approved):

config/locales/zh_CN.yml
```diff
@@ -0,0 +1,40 @@
...
+      equal_to_current_password: '必须与当前密码不同。'
+      password_complexity:
+        digit:
+          one: 必须包含至少1个数字
+          other: 必须包含至少%{count}个数字
+        lower:
+          one: 必须包含至少1个小写字母
+          other: 必须包含至少%{count}个小写字母
+        symbol:
+          one: 必须包含至少1个标点符号
```
tmr08c, line 14: Google Translate translates this to: "Must contain at least 1 punctuation mark"

In our English translations, we say punctuation mark **or symbol**.  I think symbols would include characters like "%^&*".

I'm assuming my lack of understanding is missing some nuance, but I figured I'd check if you felt this covers the ability to use punctuation and symbols.

config/locales/zh_CN.yml
```diff
@@ -0,0 +1,40 @@
...
+          other: 必须包含至少%{count}个标点符号
+        upper:
+          one: 必须包含至少1个大写字母
+          other: 必须包含至少%{count}个大写字母
+  devise:
+    invalid_captcha: '无效验证码'
```
tmr08c, line 20: Pretty sure this is another issue of me missing nuance trying to use Google Translate. Google translates this to "Invalid verification code." 

This entry deals with more with [CAPTCHAs](https://en.wikipedia.org/wiki/CAPTCHA) and not the verification code. 

Is there anything that more directly ties to CAPTCHA, or is this the best phrase? 

I'm checking because we have the idea of verification codes as a part of our `ParanoidVerificatoin` module and have some translations for them. I wanted to see if we could have them be more distinguishable if possible.

config/locales/zh_CN.yml
```diff
@@ -0,0 +1,40 @@
...
+      updated: '新密码已保存'
+      change_required: '你的密码已过期，请更新密码.'
+      show:
+        renew_your_password: 更新你的密码
+        current_password: 当前密码
+        new_password: 新密码
+        new_password_confirmation: 确认新密码
+        change_my_password: 更改密码
+    failure:
+      session_limited: '你的登录凭据已在另一个浏览器上被使用。请重新登录以在此浏览器继续。'
```
tmr08c, line 39: Is `。` the same as `.`? It looks like in some examples you use `.` and `。` in others like this.  Would it make sense to use `。` in all examples?

tmr08c (commented):
Thank you for your contribution and happy Hacktoberfest 🎃 , @dingyifei! 

I had a few clarifying questions. I am pretty sure they aren't real issues and just the result of me having to rely on Google Translate, but I figured I'd double-check.

Let me know what you think :)

config/locales/zh_CN.yml
dingyifei: The direct, word by word translation of "验证码" is verification code, but in Chinese websites, it is used for "CAPTCHAs" and "verification code". 

Since the other word for CAPTCHAs, "全自动区分计算机和人类的图灵测试" , is the direct translatin of "Completely Automated Public Turing test to tell Computers and Humans Apart", "验证码" is the best phrase for "CAPTCHAs" and "verification code". 


CAPTCHAs:
图片
Verification code:
图片

config/locales/zh_CN.yml
dingyifei: Sorry I think I made a mistake here. "标点符号" means all the characters in the set of both punchuation marks and symbols usually containing these:
图片

Chinese punchuations are included in "标点符号" but these are probably not usable in password.

In the case of only english punchuation mark and symbol, the closest phrase is "特殊符号", special symbols. The usable symbols, including english punchuations, can be clarified later with round brackets.

Could you give a list of characters that are accepted as punchuation marks and symbols? I can change it into something like "必须包含至少1个特殊符号（%^&*._）".

config/locales/zh_CN.yml
dingyifei: Yes, `。` is the Chinese period. I will change them to `。` and rephrase things I removed the periods.

config/locales/zh_CN.yml
dingyifei: I just read the `devise-security/lib/devise-security/validators/password_complexity_validator.rb` file. Punct accepts  `!"#$%&'()*+,-./:;<=>?@[\]^_‘{|}~`

https://regular-expressions.mobi/posixbrackets.html

Resolved

dingyifei:
Resolved

config/locales/zh_CN.yml
tmr08c: Thank you for explaining the options and digging into the options provided by `Punct` 🧡

config/locales/zh_CN.yml
tmr08c: Thank you for the explanation and screenshot examples 👍

tmr08c (approved):

PR #240, opened 2020-10-06
Add fa locale

mhi20:


lib/generators/devise_security/install_generator.rb
```diff
@@ -4,7 +4,7 @@ module DeviseSecurity
   module Generators
     # Generator for Rails to create or append to a Devise initializer.
     class InstallGenerator < Rails::Generators::Base
-      LOCALES = %w[en es de fr it ja tr ru nl pt].freeze
+      LOCALES = %w[en es de fr it ja tr ru nl pt fa].freeze
```
oniofchaos, line 7: Can you go ahead and alphabetize these since you're in here? Thanks 🙌

test/test_install_generator.rb
```diff
@@ -21,5 +21,6 @@ class TestInstallGenerator < Rails::Generators::TestCase
     assert_file 'config/locales/devise.security_extension.tr.yml'
     assert_file 'config/locales/devise.security_extension.nl.yml'
     assert_file 'config/locales/devise.security_extension.pt.yml'
+    assert_file 'config/locales/devise.security_extension.fa.yml'
```
oniofchaos, line 24: Can you go ahead and alphabetize these since you're in here? Thanks 🙌

oniofchaos (commented):
Thanks for adding this

lib/generators/devise_security/install_generator.rb
mhi20: Yes sure

oniofchaos (approved):
Thank you!

PR #241, opened 2020-10-08
Add Hindi (hi) locale

salap298:


config/locales/hi.yml
```diff
@@ -0,0 +1,41 @@
+
+hi:
+  errors:
+    messages:
+      taken_in_past: पहले लिया गया
```
oniofchaos, line 5: Screen Shot 2020-10-08 at 10 50 16 AM

This doesn't seem to translate quite right. The English version is `was used previously`.

config/locales/hi.yml
```diff
@@ -0,0 +1,41 @@
+
+hi:
+  errors:
+    messages:
+      taken_in_past: पहले लिया गया 
+      equal_to_current_password: इससे अलग पासवर्ड होना चाहिए
```
oniofchaos, line 6: Screen Shot 2020-10-08 at 10 51 19 AM

This seems close but not quite right. The English version is `must be different than the current password`.

config/locales/hi.yml
```diff
@@ -0,0 +1,41 @@
...
+      equal_to_current_password: इससे अलग पासवर्ड होना चाहिए 
+      password_complexity:
+        digit:
+          one: एक अंक होना चाहिए 
+          other: कम से कम %{count} अंक होने चाहिए  
+        lower:
+          one: एक लोअर-केस अक्षर होना चाहिए 
+          other: कम से कम %{count} अक्षर होने चाहिए  
+        symbol:
+          one: एक प्रतीक चिन्ह होना चाहिए
```
oniofchaos, line 15: Screen Shot 2020-10-08 at 10 52 39 AM

Don't think this is quite right. The English version is `must contain at least one punctuation mark or symbol`.

config/locales/hi.yml
```diff
@@ -0,0 +1,41 @@
...
+    invalid_security_question: सुरक्षा उत्तर गलत था 
+    paranoid_verify:
+      code_required: सपोर्ट टीम द्वारा दिया गया कोड डाले 
+    paranoid_verification_code:
+      show:
+        submit_verification_code: वेरिफिकेशन कोड डाले 
+        verification_code: वेरिफिकेशन कोड 
+        submit: सबमिट 
+    password_expired:
+      updated: नया पासवर्ड संरक्षित किया
```
oniofchaos, line 31: Screen Shot 2020-10-08 at 10 54 06 AM

Not sure if this is right? The English version is `Your new password is saved.`.

config/locales/hi.yml
```diff
@@ -0,0 +1,41 @@
...
+        verification_code: वेरिफिकेशन कोड 
+        submit: सबमिट 
+    password_expired:
+      updated: नया पासवर्ड संरक्षित किया 
+      change_required: पासवर्ड अमान्य हो चुका, पासवर्ड बदले 
+      show:
+        renew_your_password: पासवर्ड बदले 
+        current_password: अभी का पासवर्ड 
+        new_password: नया पासवर्ड 
+        new_password_confirmation: नया पासवर्ड सत्यापित
```
oniofchaos, line 37: Screen Shot 2020-10-08 at 10 55 34 AM

Not sure this is right either. The English version is `Confirm new password`.

lib/generators/devise_security/install_generator.rb
```diff
@@ -4,7 +4,7 @@ module DeviseSecurity
   module Generators
     # Generator for Rails to create or append to a Devise initializer.
     class InstallGenerator < Rails::Generators::Base
-      LOCALES = %w[en es de fr it ja tr ru nl pt uk].freeze
+      LOCALES = %w[en es de fr it ja tr ru nl pt uk hi].freeze
```
oniofchaos, line 7: Keep these alphabetized (you'll see this when you merge in latest master).

test/test_install_generator.rb
```diff
@@ -22,5 +22,6 @@ class TestInstallGenerator < Rails::Generators::TestCase
     assert_file 'config/locales/devise.security_extension.nl.yml'
     assert_file 'config/locales/devise.security_extension.pt.yml'
     assert_file 'config/locales/devise.security_extension.uk.yml'
+    assert_file 'config/locales/devise.security_extension.hi.yml'
```
oniofchaos, line 25: Keep these alphabetized (you'll see this when you merge in latest master).

oniofchaos (commented):
Note that all of the comments related to translations are discussion points. I don't speak Hindi after all and I'm relying on Google Translate here.

test/test_install_generator.rb
salap298: Ok, Thanks for the clarification/suggestion, will resolve this

config/locales/hi.yml
salap298: पहले - it can mean both first and past/previous in different contexts. See these - https://www.shabdkosh.com/search-dictionary?lc=hi&sl=en&tl=hi&e=first , https://www.shabdkosh.com/search-dictionary?lc=hi&sl=en&tl=hi&e=previous , https://www.shabdkosh.com/search-dictionary?lc=hi&sl=en&tl=hi&e=past . I will restructure the sentence.

lib/generators/devise_security/install_generator.rb
salap298: done

config/locales/hi.yml
salap298: reviewed

config/locales/hi.yml
salap298: resolved - https://www.shabdkosh.com/search-dictionary?lc=hi&sl=en&tl=hi&e=symbol

config/locales/hi.yml
salap298: resolved

salap298:
Yes I understand, but the google translations are not semantically correct most of the time.

config/locales/hi.yml
tmr08c: @salap298 thank you for making a change to this! 

I tested out adding the error to see what the full error message would look like

```ruby
self.errors.add(:password, :equal_to_current_password)
```

and it ends up as 

```
Password वर्तमान पासवर्ड के बराबर है
```

I am also going to be limited to Google Translate and miss out on the semantics, but I wanted to check. Google translates this to: 

```
Password equals current password
```

Is it possible to give more of an indication that the password should be **different** than the current password? For example, our English translation would be

```
Password must be different than the current password
```

 If not, I think we would be okay - if you get an error when updating your password that it's the same as your current that _should_ be a sign that they shouldn't match. However, if possible, I think it could be useful to be even more explicit.

config/locales/hi.yml
tmr08c: Thanks for making a change to this 🧡 

Limited to what Google is telling me, we may be losing a bit of the meaning. Google is showing the updated version, "पहले लिया जा चुका है " to be "Already taken". 

 When I read "already taken" it makes me think more like a username, like someone in the system has already used it. 

However, the goal of this error message is to say you have used this same password in the past and cannot re-use it.

As you mentioned, this is probably Google missing some of the semantics, but I thought it would be worth discussing.

config/locales/hi.yml
tmr08c: This definitely feels like a semantic/ordering thing, but I wanted to confirm.

It looks like you were able to update this to "नया पासवर्ड प्रमाणित" which Google is translating to "New password authenticate." 

My English brain wants this to be ordered the other way, more like "Authenticate new password." Is it safe to assume that this is how it would actually be read in Hindi? 

Also is there an alternative to "authenticate?" I tried the [website you shared](https://www.shabdkosh.com/search-dictionary?lc=hi&sl=en&tl=hi&e=confirm) and searched for confirm. Taking the first option, "पुष्टि करना" it seems to make Google happy, but I don't know if it _actually_ makes sense:

Screen Shot 2020-10-10 at 7 02 44 AM

tmr08c (commented):
Happy Hacktoberfest, @salap298! 🎃 👻

Thank you for making a bunch of changes already 🧡  

I re-opened a few of the threads with additional questions. As @oniofchaos pointed out, I am limited by Google Translate as well, so I appreciate you taking the time to work with us on this. I did also check out https://www.shabdkosh.com/ for one example, thank you for sharing that site :)

salap298:
Thank you

PR #242, opened 2020-10-10
Add Czech (cs) locale

CiTroNaK:


config/locales/cs.yml
```diff
@@ -0,0 +1,40 @@
...
+  devise:
+    invalid_captcha: Chybná captcha.
+    invalid_security_question: Chybná odpověď na bezpečnostní otázku.
+    paranoid_verify:
+      code_required: Zadejte kód, který poskytla naše podpora
+    paranoid_verification_code:
+      show:
+        submit_verification_code: Odeslat ověřovací kód
+        verification_code: Ověřovací kód
+        submit: Odeslat
```
oniofchaos, line 28: Not sure if the translation is right here? It's coming across as "Send" on Google Translate, but the correct word is "Submit".

Screen Shot 2020-10-15 at 10 17 45 AM

Screen Shot 2020-10-15 at 10 17 32 AM

test/test_install_generator.rb
```diff
@@ -26,5 +26,6 @@ class TestInstallGenerator < Rails::Generators::TestCase
     assert_file 'config/locales/devise.security_extension.ru.yml'
     assert_file 'config/locales/devise.security_extension.tr.yml'
     assert_file 'config/locales/devise.security_extension.uk.yml'
+    assert_file 'config/locales/devise.security_extension.cs.yml'
```
oniofchaos, line 29: Alphabetize this within the rest of the files please!

lib/generators/devise_security/install_generator.rb
```diff
@@ -3,8 +3,9 @@
 module DeviseSecurity
   module Generators
     # Generator for Rails to create or append to a Devise initializer.
+
```
oniofchaos, line 6:
```suggestion
```

oniofchaos (commented):
Overall looks great! Just 2 nitpicks and a translation question.

config/locales/cs.yml
CiTroNaK: @oniofchaos This is a common word for this kind of action. The "předložit" or similar is used more like for official communication (between companies, state administration, etc).

oniofchaos: Ahhhhh that makes a lot of sense. Thanks!

oniofchaos (approved):

oniofchaos (approved):

PR #243, opened 2020-10-10
Allow random failures for unsupported Ruby/Rails pairs

dlackty:
While investigating random CI failures for this project, I noticed that this proejct is tested on Ruby 2.7.1 with Rails < 6, which is officially unsupported  (rails/rails#38426).

I think we should still add these version combinations to Travis CI's allow failures list, and hope this can make CI more stable.


oniofchaos (approved):
epic, thanks

PR #244, opened 2020-10-10
Add zh_TW translation

dlackty:
As titled, thank you for making this nice library.


config/locales/zh_TW.yml
```diff
@@ -0,0 +1,40 @@
+zh_TW:
+  errors:
+    messages:
+      taken_in_past: '曾被使用過。'
```
oniofchaos, line 4: Not sure if we want the extra space at the end?

config/locales/zh_TW.yml
```diff
@@ -0,0 +1,40 @@
...
+          one: 必須包含至少一個小寫字母
+          other: 必須包含至少 %{count} 個小寫字母
+        symbol:
+          one: 必須包含至少一個特殊符號
+          other: 必須包含至少 %{count} 個特殊符號
+        upper:
+          one: 必須包含至少一個大寫字母
+          other: 必須包含至少 %{count} 個大寫字母
+  devise:
+    invalid_captcha: '輸入的驗證碼無效。'
```
oniofchaos, line 20: Screen Shot 2020-10-20 at 4 05 03 PM

Not sure if this is right? We aren't entering a verification code here.

dlackty: Verification code & recaptcha are actually translated into same Traditional Chinese word.

CleanShot 2020-10-21 at 12 28 08@2x

You can also find [Wikipedia page](https://zh.wikipedia.org/zh-tw/%E9%AA%8C%E8%AF%81%E7%A0%81) using this translation.

config/locales/zh_TW.yml
dlackty: Sorry if I'm wrong but I checked with different text editors and couldn't find the extra space?

oniofchaos: Oh interesting. It looks like there's space between it and the single quote at the end, but I guess that's just part of whatever character that's called? Anyway, verified through a text editor and there's no space character here so that's fine.

oniofchaos (approved):

Issue #238, opened 2020-10-06
Rubocop warning on initializer file name


DevS1993:
So i recently added this extension to simplify my password validation flow and i noticed this error from Rubucop.

![devise_security_rubocop](https://user-images.githubusercontent.com/21246070/95175297-ea4a4b80-07d8-11eb-816d-da196f369159.png)

Other initializers (like devise_token_auth.rb / filter_parameter_logging.rb) follow snake_case file naming convention.

is the file name `devise-security.rb` intentionally named like this and i have to add an exception to Rubocop config?

olbrich:
I'm pretty sure that rails will load it just fine if you rename the file. Let us know if you have trouble. I would also welcome a PR to make the name more consistent with rubocop.

DevS1993:
Yes @olbrich rails loads it even after changing the file name follow snake_case. No troubles. I'll work on a PR to change the default behavior.

tmr08c:
@DevS1993 I've assigned the Issue to you. Let me know if you would like me to unassign it.

🤔  In the short-term, you may be able to rename the file to `devise_security.rb` in your application. I think the important part is that it will still run `Devise.setup` and let you configure anything you need. The actual file name _shouldn't_ matter.

PR #245, opened 2020-10-10
Tweak initializer file name to remove rubocop warning (#238)

DevS1993:
* Change name to remove rubocop warning


oniofchaos:
What's the rubocop warning that this was causing?

DevS1993:
@oniofchaos thisi the warning:

devise_security_rubocop

I made the changes as per discussions in #238 .

lib/generators/devise_security/install_generator.rb
```diff
@@ -11,8 +11,8 @@ class InstallGenerator < Rails::Generators::Base
       desc 'Install the devise security extension'
 
       def copy_initializer
-        template('devise-security.rb',
```
oniofchaos, line 14: We will also need to support the non-snake-case version because:

1) Support for existing codebases so that this isn't a breaking change.
2) For anyone who disagrees with the cop and wants the dash instead of the underscore.

README.md
```diff
@@ -54,7 +54,7 @@ rails generate devise_security:install
 ```
 
 The generator adds optional configurations to
-`config/initializers/devise-security.rb`. Enable the modules you wish to use in
```
oniofchaos, line 57: Per my other comment, let's add a note here that `devise-security` is supported but doesn't match Rubocop's style guide, with a link to https://docs.rubocop.org/rubocop/cops_naming.html#namingfilename

oniofchaos (commented):
I'm all for it, just two things to change.

lib/generators/devise_security/install_generator.rb
tmr08c: @oniofchaos by support, are you thinking we should be able to generate both filenames? 

From my understanding of how things work, it seems like the file name doesn't end up really mattering. I believe Rails will load all files in the `config/initializers` directory and I don't think the code has any direct reference to the `devise-security.rb`, we just expect the config file to leverage `Devise.setup`:

https://github.com/devise-security/devise-security/blob/ad50cdb888473c38b4023c2343f6c014604f5b84/lib/generators/templates/devise-security.rb#L3

oniofchaos: That's a very good point. I misread what some of this code was doing.

One thing I'd like to see before this is merged is an example app that has the "old" config name and uses this branch of the gem and demos that it still works.

olbrich (approved):

natebird (approved):

oniofchaos:
@DevS1993 tried fixing merge conflicts through the browser but Github didn't like what I did. Can you merge in latest master to resolve them (reverting my commits if you need to) so I can merge?

PR #248, opened 2020-10-15
Reduce gem size

igas:
I'm working on docker optimisations for our app and trying to reduce the size of the gems we are using. Gem does not need to contain all of the files, so I reduced it. `test` folder is more than half of the whole gem.


oniofchaos (approved):
Brilliant!

PR #249, opened 2020-10-15
Bump default dev ruby to 2.7.1

igas:
While I was doing #248 rvm complained about ruby version. Because CI supports 2.7, I thought it might be a good idea to bump `.ruby-version` too.


oniofchaos (approved):
Works for me!

olbrich:
Actually, I'm going to have to revert this, the idea behind our maintenance is to keep the development stack linked to the oldest supported version of ruby so we don't try to make use of new language features and introduce compatibility bugs as a result.

igas:
@olbrich do you think CI would be too late for feedback loop? I can see a point in it, but even then, it should be `2.3.0` as per gemspec.

PR #254, opened 2020-10-29
Improve test coverage for password expirable

oniofchaos:
`params.permit` was added in Rails 4.0, so we don't need to keep the legacy check.


test/controllers/test_password_expired_controller.rb
```diff
@@ -21,11 +21,35 @@ class Devise::PasswordExpiredControllerTest < ActionController::TestCase
...
+    assert_redirected_to :root
+  end
+
+  test 'redirects on show if user does not need password change' do
+    @user.update(password_changed_at: Time.zone.now)
+    get :show
+    assert_redirected_to :root
+  end
+
   test 'should render show' do
```
tmr08c, line 36: Does it make sense to add more context to match the other `test` strings?

```suggestion
  test 'should render show when user needs a password change' do
```

tmr08c (approved):
Out of curiosity, what inspired adding these tests? Did any issues come up?

olbrich (approved):

PR #253, opened 2020-10-29
Bump devise-security to 0.15.0

oniofchaos:

PR #260, opened 2020-12-22
Separate positional and keyword arguments

hnatt:
Ruby 3.0 will [separate positional and keyword arguments][1], so Ruby 2.7 gives deprecation warnings about code that will become incompatible:

```
devise-security/lib/devise-security/controllers/helpers.rb:91: warning: Using the last argument as keyword parameters is deprecated; maybe ** should be added to the call
```

This PR fixes such warnings.

[1]: https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/


oniofchaos (approved):
LGTM will merge once https://github.com/devise-security/devise-security/pull/253 is released.

tmr08c (approved):
Thank you for your contribution and useful write-up, @hnatt 🎉 ! 

I confirmed that I see the warning when running the tests using Ruby 2.7.1 and it goes away on this branch 👍 

The docs for i18n also confirm they are using keyword arguments over hashes:

https://github.com/ruby-i18n/i18n/blob/4709391dceab9096d5988576f93935843023a6ef/lib/i18n.rb#L180-L193

PR #261, opened 2020-12-31
Update devise-security.gemspec

oniofchaos:
Ensure required ruby version matches what we officially support. Follow up to https://github.com/devise-security/devise-security/pull/189.

### DIFF
--- a/devise-security.gemspec
+++ b/devise-security.gemspec
@@ -25,7 +25,7 @@ Gem::Specification.new do |s|
   s.files         = Dir['README.md', 'LICENSE.txt', 'lib/**/*', 'app/**/*', 'config/**/*']
   s.test_files    = Dir['test/**/*']
   s.require_paths = ['lib']
-  s.required_ruby_version = '>= 2.3.0'
+  s.required_ruby_version = '>= 2.5.0'
 
   s.add_runtime_dependency 'devise', '>= 4.3.0', '< 5.0'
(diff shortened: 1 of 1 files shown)


olbrich:
There is a difference between a required ruby version and a supported one. I do not think we should bump the required ruby version unless we start using a language feature that is not supported by an older version.  If we update it then anyone using this gem on ruby 2.4 will not be able to install and use it even if all we did was stop supporting that version.  I think we should abandon this change unless there is a known incompatibility with 2.3 or 2.4.

oniofchaos:
So the concern I had is that we haven't been running tests since April so there very well could be an incompatibility that we won't find out about until we push out the new version. I wasn't sure how to resolve that except by requiring a newer version.

olbrich:
We don't support the old versions, but we also don't want to deliberately break it for anyone who might be using them either. Making this change would move us from 'might not work on older ruby versions' to "won't even install in older versions".  I'd rather not force that unless we know for sure that it's a problem.  We aren't going to run tests against the older versions to detect it since we aren't supporting them anymore, but if someone using an older version were to report a bug indicating that it's not compatible with 2.3 or 2.4, then my response would be to require a newer ruby version instead of trying to fix the bug.

oniofchaos:
That makes a lot of sense to me. I like it. Thanks for the input.

PR #270, opened 2021-01-20
Allow combinations of lower or upper and symbol or digit.

jackslinger:
Please note that this is my first PR, so I may have made some mistakes. Please let me know if there is anything I can change to make this PR easier to review.

## Reason for change

A project I have been working on has a requirement to allow passwords with at least one letter (upper or lowercase) and at least one (symbol or digit). So `password1234` should be a valid (if bad) password and so should `password!-_@`, but `password` should not be.

To support this I have added two new keys to the password complexity validator to allow lower_or_upper, and symbol_or_digit.

## Alternatives considered

We considered using a custom validation and writing our own regex for the password validation. In the end we decided against this as the `devise-security` interface is nicer to work with and easier to read.


jackslinger:
I think these tests are failing tests are fixed under #268?

README.md
```diff
@@ -89,7 +89,7 @@ Devise.setup do |config|
   # config.expire_password_after = 3.months | true | false
 
   # Need 1 char each of: A-Z, a-z, 0-9, and a punctuation mark or symbol
-  # config.password_complexity = { digit: 1, lower: 1, symbol: 1, upper: 1 }
```
natebird, line 92: I think I would like to see what happens if all of the complexity parameters are passed. Some explanation might be necessary to explain the side effects.

oniofchaos: Agreed

olbrich (commented):
Looks like the comments in https://github.com/devise-security/devise-security/pull/270/files#diff-1620e28c960dc90da1115b1a6bc97b4e170e99a257e29addf037eed8f33aff4e need to be updated and there are some translations that need to be written.

tmr08c (commented):
Thank you for your first contribution, @jackslinger 🎉  !

In addition to the comments from others, I had a small suggestion for one of the tests.

Also, I believe @olbrich was referencing 

https://github.com/devise-security/devise-security/blob/17b85439e3718a4beca6a29544b34cfa0b310177/config/locales/en.yml#L7-L19

hopefully, that helps make it a bit easier to update :)

olbrich:
A couple of things...

1. NIST now no longer recommends complexity rules for passwords as it just makes it more likely that people will work around them with easy to guess algorithms... think `password` vs. `password1`.  That said, people still have to deal with requirements from their security teams that haven't adapted to that guidance.
2. The current design only covers a few cases and doesn't really allow much flexibility.  It should be possible to create a custom password complexity validator that subclasses and extends the one we provide, and maybe some documentation about how to do that would be in order.
3. The current design does not check for patterns that should be excluded (e.g., long sets of repeated characters, excluded characters, etc..).  (I might be persuaded to refactor the validator to cover this case, but even this rule is deceptively complex to implement correctly).

Given that NIST no longer recommends this approach (see https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-63b.pdf section A.3) my inclination is to not update the functionality for corner cases like this that aren't generally useful and instead provide guidance (and possibly update code) to make it easy for users to replace or extend the password complexity validator.

@dillonwelch @natebird thoughts?

dillonwelch:
Yup, I like the idea of adding the ability to add a custom validator and leaving it at that. Thinking we should close this PR, make that an issue, and work on it for the next version. What do you think?

dillonwelch:
Talked more with @olbrich and we are going to go down the custom validator approach. I made https://github.com/devise-security/devise-security/issues/360 to track discussion and work on that idea.

PR #277, opened 2021-02-10
Add validation to ensure password does not match email to secure_validatable

arocca1:
This simple validation protects against a very simple password which an
attacker would be able to fairly simply guess or attempt. It's a bad
practice, so let's make sure that it's not possible to set such a
password :).

This PR adds in an additional Devise config variable `allow_passwords_equal_to_email`
which can be used to disable the password check, if desired. By default, passwords
cannot equal the email (in any case sensitive way), i.e. the config variable is false.


arocca1:
I suspect that at least bumping `database_cleaner` (https://github.com/devise-security/devise-security/pull/275) will fix my test failures. The newer versions of `mongoid` (5.0+) are failing with:
```
/home/runner/work/devise-security/devise-security/gemfiles/vendor/bundle/gems/database_cleaner-core-2.0.1/lib/database_cleaner/cleaner.rb:93:in `rescue in orm_strategy': The 'truncation' strategy does not exist for the mongoid ORM!  Available strategies:  (DatabaseCleaner::UnknownStrategySpecified)
	from /home/runner/work/devise-security/devise-security/gemfiles/vendor/bundle/gems/database_cleaner-core-2.0.1/lib/database_cleaner/cleaner.rb:88:in `orm_strategy'
	from /home/runner/work/devise-security/devise-security/gemfiles/vendor/bundle/gems/database_cleaner-core-2.0.1/lib/database_cleaner/cleaner.rb:85:in `create_strategy'
	from /home/runner/work/devise-security/devise-security/gemfiles/vendor/bundle/gems/database_cleaner-core-2.0.1/lib/database_cleaner/cleaner.rb:45:in `strategy='
	from /home/runner/work/devise-security/devise-security/test/orm/mongoid.rb:11:in `<top (required)>'
	from /home/runner/work/devise-security/devise-security/gemfiles/vendor/bundle/gems/activesupport-5.0.7.2/lib/active_support/dependencies.rb:293:in `require'
	from /home/runner/work/devise-security/devise-security/gemfiles/vendor/bundle/gems/activesupport-5.0.7.2/lib/active_support/dependencies.rb:293:in `block in require'
	from /home/runner/work/devise-security/devise-security/gemfiles/vendor/bundle/gems/activesupport-5.0.7.2/lib/active_support/dependencies.rb:259:in `load_dependency'
	from /home/runner/work/devise-security/devise-security/gemfiles/vendor/bundle/gems/activesupport-5.0.7.2/lib/active_support/dependencies.rb:293:in `require'
	from /home/runner/work/devise-security/devise-security/test/test_helper.rb:35:in `<top (required)>'
	from /home/runner/work/devise-security/devise-security/test/test_compatibility.rb:1:in `require'
	from /home/runner/work/devise-security/devise-security/test/test_compatibility.rb:1:in `<top (required)>'
	from /home/runner/work/devise-security/devise-security/gemfiles/vendor/bundle/gems/rake-13.0.3/lib/rake/rake_test_loader.rb:17:in `require'
	from /home/runner/work/devise-security/devise-security/gemfiles/vendor/bundle/gems/rake-13.0.3/lib/rake/rake_test_loader.rb:17:in `block in <main>'
	from /home/runner/work/devise-security/devise-security/gemfiles/vendor/bundle/gems/rake-13.0.3/lib/rake/rake_test_loader.rb:5:in `select'
	from /home/runner/work/devise-security/devise-security/gemfiles/vendor/bundle/gems/rake-13.0.3/lib/rake/rake_test_loader.rb:5:in `<main>'
rake aborted!
```

natebird:
I'm working on a dependency update that should resolve that. See #278.

arocca1:
Oh, great, thanks!

lib/devise-security/models/secure_validatable.rb
```diff
@@ -70,6 +73,15 @@ def current_equal_password_validation
         self.errors.add(:password, :equal_to_current_password) if dummy.valid_password?(password)
       end
 
+      def email_not_equal_password_validation
+        return if password.blank? || (!new_record? && !will_save_change_to_encrypted_password?)
+        dummy = self.class.new.tap do |user|
+          user.password_salt = password_salt if respond_to?(:password_salt)
+          user.password = email
```
arocca1, line 80: Upon further consideration, this requires further attention to potential uses of `case_insensitive_keys`. If email is used as case insensitive, then this validation will not be semantically correct. It would only fail validation if the password exactly matches the downcased email.
An alternative here could be to downcase `password` if `case_insensitive_keys` is set

arocca1:
It's worth noting that I explicitly didn't do any sort of `include`-like check because that is too restrictive and doesn't exactly match the intention of this change. The goal here is to protect against a user trying to set a password that matches their email (or something that may seem "clever", like just adding spaces around it). As a result, checking that a password includes the email in any form is out of scope.
I'm open to further discussion here though if it should be in scope. I can appreciate that a password of `me@email.com1` for an email of `me@email.com` is easily guessable too.

arocca1:
I also recognize that having such a default is fairly restrictive and generally odd with respect to backwards compatibility, but having a password match the email is a really easily guessable password and I believe it to be worth protecting against by default.

lib/generators/templates/devise-security.rb
```diff
@@ -41,4 +41,10 @@
 
   # Time period for account expiry from last_activity_at
   # config.expire_after = 90.days
+
+  # Allow passwords to be equal to email (false, true)
+  # This specifically performs case insensitive comparisons
+  # Used by secure_validatable. Defaults to false
+  # config.allow_passwords_equal_to_email = false # deny passwords equal (case insensitive) to the email
```
oniofchaos, line 48: Let's only have one line, the user can change the value as they want (similar to the other configs in this file).

oniofchaos (commented):
Everything looks good besides the one config thing.

lib/generators/templates/devise-security.rb
arocca1: Sounds good. Thanks for looking. I'll push that up in a few minutes

oniofchaos:
I made another change to master, merge that in and it's good to go.

arocca1:
Thanks!

dillonwelch:
Hi @tiagotex, thanks for bringing this up. I created https://github.com/devise-security/devise-security/issues/349 to move the discussion.

PR #286, opened 2021-03-01
Avoid any global sort orders

experian-greg-myers:
For anyone using UUIDs as default in system, use reorder rather than order to avoid chained orders from taking effect.


oniofchaos:
Can you add a test using UUIDs to demonstrate that this works + prevent regressions?

experian-greg-myers:
Will do. Accidentally put this PR under work account instead of home account so will also close and re-open PR under the correct account this evening.

experian-greg-myers:
Closed, #289 has tests.

Issue #290, opened 2021-03-18
:secure_validatable doesn't set @minimum_password_length


dredupuika:
devise :validatable does this
```# Sets minimum password length to show to user
def set_minimum_password_length
  if devise_mapping.validatable?
    @minimum_password_length = resource_class.password_length.min
  end
end
```

oniofchaos:
Hi @dredupuika. I'm not sure what your issue is about. Is it a bug report, a feature request, or something else?

dredupuika:
Obviously a bug. 
Since :secure_validatable is a replacement for the original :validatable it should not brake any functionality.
In this case it makes the original forms in devise less user friendly. As a result 2 devise controllers must be overriden to fix this.

It should at least be mentioned in the readme.

oniofchaos:
Can you share with me what controllers / methods you override in order to fix this?

dredupuika:
I haven't finished and tested it thoroughly. And I consider it a temporary fix.
But since devise uses [set_minimum_password_length](https://github.com/heartcombo/devise/search?q=set_minimum_password_length) in 2 controllers. So I made custom controllers for **registrations** and **passwords** with this method:
```
def set_minimum_password_length
  if devise_mapping.secure_validatable?
    @minimum_password_length = resource_class.password_length.min
    @minimum_password_complexity = resource_class.password_complexity
  else
    super
  end
end
```

however i do think that this problem should be addressed by overriding the devise_controller directly.

oniofchaos:
Ok great! I have some ideas on how to address this. I'll post back once I've been able to explore it.

oniofchaos:
Can you test with https://github.com/oniofchaos/devise-security/tree/validatable and see if it solves the problem? You shouldn't need any controller overrides anymore on either controller.

dredupuika:
Yeah. Works as expected.

I will be keeping those controllers for now since they expose the complexity requirements to the form in the same manner as devise does.
But that's probably feature for the future.

Thanks.

oniofchaos:
Can you elaborate on how you're using the exposed complexity requirements? I'd be happy to add those too.

dredupuika:
After this fix.
In the controller:
```
def set_minimum_password_length
  @minimum_password_complexity = resource_class.password_complexity if devise_mapping.secure_validatable?
  super
end
```

In the form: (I use the bootstrap versions)
```
# original
- if @minimum_password_length
[...]

PR #291, opened 2021-03-26
Ensure @minimum_password_length is set for devise views when using secure_validatable

dillonwelch:
See https://github.com/devise-security/devise-security/issues/290


lib/devise-security/patches/secure_validatable_information.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    included do
+      alias_method(
+        :original_set_minimum_password_length, :set_minimum_password_length
+      )
+
+      define_method :set_minimum_password_length do
+        if devise_mapping.secure_validatable?
+          @minimum_password_length = resource_class.password_length.min
+          @minimum_password_complexity = resource_class.password_complexity
```
olbrich, line 15: I'm not sure why we would need to expose this?  The `@minimum_password_length` is used in some of the default views, so I can understand that, but the password complexity is not.

Isn't it just easier to directly access the configurations for `password_length` and `password_complexity` in custom views?

oniofchaos: The idea from https://github.com/devise-security/devise-security/issues/290 was for the same reason that password length is exposed for views, if password complexity was exposed then people could default use them in their views in the same fashion. 

Do you think it would be better to just support `@minimum_password_length` for devise purposes and then note if you want to access password complexity in your views that you should use the config directly instead?

dredupuika: Let me just add that while i do agree that it is easier to access `password_length` and `password_complexity` in the views directly. However in this scenario the views also need to check if `secure_validatable` is used.

This is why I did prefer to use the same approach as devise does. Just to keep the inner workings out of the views.

oniofchaos: @olbrich bump on your opinions for this?

Issue #267, opened 2021-01-13
Investigate updating devise past 4.7.3 once it's released


oniofchaos:
This version has a soft-lock of using 1.x.

From https://github.com/devise-security/devise-security/runs/1698408739?check_suite_focus=true

> /home/runner/work/devise-security/devise-security/gemfiles/vendor/bundle/gems/devise-4.7.3/lib/devise/omniauth.rb:12:in `<top (required)>': You are using an old OmniAuth version, please ensure you have 1.0.0.pr2 version or later installed. (RuntimeError)

Devise issue where this is being discussed: https://github.com/heartcombo/devise/issues/5326

natebird:
The latest update in the thread above says Devise with support for omniauth 2.0.0+ will be sometime in February.

olbrich:
Devise released 4.8.0 a few days ago.

PR #313, opened 2021-05-04
Remove lock on omniauth since devise is now on 4.8.0

oniofchaos:
Closes https://github.com/devise-security/devise-security/issues/267


olbrich:
@oniofchaos I think you are going to want to run `bundle exec appraisal generate` and check in any changes to the gemfiles.

oniofchaos:
When I run it locally, there are no additional changes. Are you seeing something different locally?

Issue #311, opened 2021-05-04
Audit what gems of Rails we depend on and add them as dependencies


oniofchaos:
With https://github.com/devise-security/devise-security/releases/tag/v0.16.0, we have removed support for Rails 4.2. However, we do not have any direct dependencies on Rails or any of its gems. We get a dependency on `railties` thanks to `devise`, but for the version we are on are very permissive and allow 4.1+. This would technically allow someone on Rails 4.2 to update to 0.16.0 and have a gem that's not functional.

In local development, we pull in all of Rails due to the `easy-captcha` gem. This has masked the above issue. 

We should explicitly state which gems of Rails we depend on and ensure their version requirements match what we test for.

olbrich:
Well, `activerecord` for sure, and probably `activesupport`.

PR #315, opened 2021-05-06
Add Rails as an explicit dependency

oniofchaos:
Closes https://github.com/devise-security/devise-security/issues/311

The testing architecture is dependent on Rails being present, so it was extremely hard to figure out how to include just the required sub-dependencies and ensure that the library still works.

We would have to rewrite the tests to 
A) not require a Rails dummy application 
B) manually load things like ActiveRecord database pooling. 

It's definitely possible, but a much larger project than I was anticipating.

Therefore, I decided to go ahead and require the entirety of Rails.

### DIFF
--- a/Appraisals
+++ b/Appraisals
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 appraise 'rails-5.0' do
-  gem 'railties', '~> 5.0.0'
+  gem 'rails', '~> 5.0.0'
   gem 'minitest-rails', '~> 5.0.0'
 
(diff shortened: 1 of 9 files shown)


olbrich:
https://github.com/devise-security/devise-security/pull/219

We had this as a dependency before and explicitly removed it.

oniofchaos:
Thanks for reminding me...especially because I merged it...

What do you think about setting `railties` as a dependency, since that's already required by devise (just at a less restrictive version)

Issue #144, opened 2019-10-28
Customize redirect route


MasterBroki:
Hi,

Is there a way to customize the redirection for expired password or deactivate it?

If so, could we document it a bit? thanks

MasterBroki:
After some research in the source code, I need to override the redirect_for_password_change(scope) method in the base controller. Can we add this in the wiki so it helps someone else at some point?

markblasdls:
@MasterBroki Hi, can you share how did you solve it with redirect_for_password_change? I'm currently having problem with redirect and can't override it. Thankyou in advance.

MasterBroki:
@markblasdls  overriding it didn't work?

markblasdls:
@MasterBroki yeah it didn't work for me but was able to find other way by modifying store_location_for(scope) and it solve my problem.. btw thankyou for your response.

MasterBroki:
Maybe you're on a different version. Anyway, this needs more doc

PR #316, opened 2021-05-06
Add ability to customize redirect routes for password expire / paranoid

dillonwelch:
Closes https://github.com/devise-security/devise-security/issues/144


app/controllers/devise/paranoid_verification_code_controller.rb
```diff
@@ -41,4 +41,10 @@ def authenticate_scope!
     send(:"authenticate_#{resource_name}!")
     self.resource = send("current_#{resource_name}")
   end
+
+  def redirect_route
+    Devise.paranoid_verification_code_redirect_location ||
+      stored_location_for(scope) ||
+      :root
+  end
```
olbrich, line 49: Perhaps we want to respect the stored location first and then fall back to the configured default.

```suggestion
  # Where to redirect to follow successful verification of the code
  # @return [String, Symbol]
  def redirect_route
    stored_location_for(scope) || Devise.paranoid_verification_code_redirect_location
  end
```

app/controllers/devise/password_expired_controller.rb
```diff
@@ -52,4 +52,10 @@ def authenticate_scope!
     send(:"authenticate_#{resource_name}!")
     self.resource = send("current_#{resource_name}")
   end
+
+  def redirect_route
+    Devise.password_expired_redirect_location ||
+      stored_location_for(scope) ||
+      :root
+  end
```
olbrich, line 60:
```suggestion
  # Where to redirect to after successfully updating the password
  # @return [String, Symbol]
  def redirect_route
    stored_location_for(scope) || Devise.password_expired_redirect_location
  end
```

lib/devise-security.rb
```diff
@@ -83,6 +83,14 @@ module Devise
   # Whether to allow passwords that are equal (case insensitive) to the email
   mattr_accessor :allow_passwords_equal_to_email
   @@allow_passwords_equal_to_email = false
+
+  # Custom redirect route for successful password expiration update action
+  mattr_accessor :password_expired_redirect_location
+  @@password_expired_redirect_location = nil
```
olbrich, line 89: We can just set the default here instead of in the controllers

```suggestion
  @@password_expired_redirect_location = :root
```

lib/devise-security.rb
```diff
@@ -83,6 +83,14 @@ module Devise
...
+
+  # Custom redirect route for successful paranoid verification code update action
+  mattr_accessor :paranoid_verification_code_redirect_location
+  @@paranoid_verification_code_redirect_location = nil
```
olbrich, line 93:
```suggestion
  @@paranoid_verification_code_redirect_location = :root
```

olbrich:
@dillonwelch You are also going to want to update the lib/generators/templates/devise_security.rb file to include the relevant configuration documentation.

app/controllers/devise/paranoid_verification_code_controller.rb
dillonwelch: The idea behind this setting was that it was an override setting, not a configured default setting. That's also why I defined it as `nil`, so that if it wasn't set it would just use the pre-existing logic.

olbrich: My suggestion should be equivalent except that it pulls the definition for what the default is into the configuration code instead of leaving it buried in the controllers.

olbrich: The biggest difference here is that the configuration you propose would completely ignore any 'stored location'. I don't think that's what we want, however.

dillonwelch: I think it is - the stored location is always set by the controller, so if we put it first on the priority list then the new config will never be used.

See https://github.com/devise-security/devise-security/blob/2b8a56884125dae487f1368cba42fec7e78b919a/lib/devise-security/controllers/helpers.rb#L64

olbrich:
@dillonwelch After looking at the Devise code a bit more it appears that the way they handle custom redirect paths is to define a method in controllers named something like `after_resetting_password_path_for(resource)` and then recommending the users override that method to do whatever custom  routing is needed.  If we take that approach then we should just define and document reasonably named methods and not add additional configuration options.

dillonwelch:
Oh that sounds much better than the config options approach. I'll take a look into that.

config/locales/en.yml
```diff
@@ -23,6 +23,7 @@ en:
     paranoid_verify:
       code_required: 'Please enter the code our support team provided'
     paranoid_verification_code:
+      updated: Verification code accepted
```
olbrich, line 26: This will probably raise an error in non-english localizations. Contributions are welcome.

olbrich (approved):

PR #320, opened 2021-05-18
FIX `Devise::Models::SecureValidatable#email_not_equal_password_validation`

RDeckard:
`Devise::Models::SecureValidatable#email_not_equal_password_validation` crashes with a `nil` email, whereas in case of blank email this validation should be skip (password validation context)


oniofchaos:
Fix looks good. Please add a test to cover this case!

oniofchaos:
Hmm looks like there's a problem with the library + ruby head. I'll look into it.

PR #321, opened 2021-07-06
Add default_scoped to prevent deprecation exception

llopez:
DEPRECATION WARNING: Class level methods will no longer inherit scoping from `create!` in Rails 6.1. To continue using the scoped relation, pass it into the block directly. To instead access the full set of models, as Rails 6.1 will, use `User.default_scoped`.


oniofchaos:
Looks like there are some failures on certain combos (mainly older versions it looks like). You will need to add some kind of check to use the supported method for each version. 

You can ignore the `LoadError: cannot load such file -- net/smtp` errors as that's a separate issue I'm trying to fix.

llopez:
I need to investigate a little more. I guess what I did is wrong. There are other parts in the code with the same problem. I will close it until I have a better understanding of the problem. sorry

PR #327, opened 2021-10-10
Adds Bulgarian localization

kirichkov:
Adds Bulgarian translation of the locales


oniofchaos:
The test suite should be fixed now. Please merge in latest master and this is good to go.

PR #345, opened 2021-11-16
Fix translations for digits/symbols keys on password complexity

dillonwelch:
The following options should add the same error message:
* `{ digit: 1 }` and `{ digits: 1 }`
* `{ digit: 2 }` and `{ digits: 2 }`
* `{ symbol: 1 }` and `{ symbols: 1 }`
* `{ symbol: 2 }` and `{ symbols: 2 }`

As there were no translations for the `digits` and `symbols` keys by
default, this would result in a missing translation error.

Instead of having to add a separate locale key for `digits` and
`symbols`, I opted to map those keys to the correspoding singular word.
This way, as long as the singular word is translated, the plural word
will be too with no extra effort.

It could be argued that `{ digit: 2 }` and `{ symbol: 2 }` should not
be allowed due to them being singular words with a plural count. I
chose not to go down this route for two reasons:
* It would add additional error handling code to maintain as well as
  throw an exception for something that otherwise would not (unless
  the user enabled a specific setting).
* Based on the amount of contributed translation files, there are many
  non-native English speakers using the gem. This is a minor enough
  issue that they shouldn't be punished for it.
  
Separated out this change from 
https://github.com/devise-security/devise-security/pull/63.

Indirectly part of 
https://github.com/devise-security/devise-security/issues/62

Co-authored-by: Kevin Olbrich <kevin.olbrich@gmail.com>


lib/devise-security/validators/password_complexity_validator.rb
```diff
@@ -1,37 +1,56 @@
...
 
-  def active_pattern_keys
-    options.keys & PATTERNS.keys
+  # A Hash of the valid patterns that have been enabled. These are normalized to
+  # have singular key names and positive limits. Negative or zero limits are
+  # ignored.
+  #
+  # @return [Hash<Symbol,Integer>]
+  def active_patterns
+    options.transform_keys { |pattern| pattern.to_s.singularize.to_sym } # singularize keys
```
dillonwelch, line 51: I think you can make this a lot cleaner with a regular loop and a few extra lines of code instead of chained pretty methods. It should also condense this from 4 loops into 1 (not that performance is a huge deal with a small constant size array...)

```ruby
    {}.tap do |active_patterns|
      options.each do |option, value|
        normalized_option = option.to_s.singularize.to_sym
        next unless patterns.key?(normalized_option)

        transformed_value = [0, value.to_i].max # I think it's possible we could just do value.to_i.positive? on the next line
        active_patterns[normalized_option] = transformed_value if transformed_value.positive?
      end
    end
```

olbrich: I took this idea and used it to simplify the validator even more

Issue #349, opened 2021-11-19
allow_passwords_equal_to_email config greatly slows down User validation


dillonwelch:
See https://github.com/devise-security/devise-security/pull/277#issuecomment-973099879

dillonwelch:
@tiagotex I was able to reproduce the slowdown you noticed. It looks like it is related to the strategy used of creating a duplicated User model and setting the password on it, an operation that takes some time. I'm looking into how we can do this validation without going through that process.

dillonwelch:
@tiagotex I do see significant improvement in the test suite when I use an alternative method of validating the password (from 92s to 56s).

Can you try out this branch and see if it fixes the issue for you? https://github.com/dillonwelch/devise-security/tree/fix-user-perf

tiagotex:
Hi @dillonwelch 

Thank you so much for taking the time to address this 💚

I will test it tomorrow and will share the results here.

tiagotex:
@dillonwelch Can confirm this fixes the perfomance regression 🙌

dillonwelch:
Great, will work in polishing off when I can.

PR #354, opened 2021-12-20
Refactor email/password check for performance

dillonwelch:
Closes https://github.com/devise-security/devise-security/issues/349

Previous method relied on the `valid_password?` method that is
purposefully non-performance. In this scenario, we do not need to
validate the passsword - only that the contents are not equal to the
user's email address.


lib/devise-security/models/secure_validatable.rb
```diff
@@ -76,14 +76,11 @@ def current_equal_password_validation
...
 
       def email_not_equal_password_validation
+        return if self.class.allow_passwords_equal_to_email
         return if password.blank? || email.blank? || (!new_record? && !will_save_change_to_encrypted_password?)
-        dummy = self.class.new.tap do |user|
-          user.password_salt = password_salt if respond_to?(:password_salt)
-          # whether case_insensitive_keys or strip_whitespace_keys include email or not, any
-          # variation of the email should not be a supported password
-          user.password = email.downcase.strip
+        if ActiveSupport::SecurityUtils.secure_compare(password.downcase.strip, email.downcase.strip)
```
dillonwelch, line 81: @olbrich note that I ended up using `secure_compare` and not `fixed_length_secure_compare` because `fixed_length_secure_compare` raises an exception if the lengths aren't the same whereas `secure_compare` returns `false` (which is what we want).

dillonwelch:
Re-ran the test suite a couple times after https://github.com/devise-security/devise-security/pull/354/commits/aacaa6d939cc74318710698b383ec1de963d725a and that seems to have cleared the flakiness up for the moment. I'll keep an eye on it and adjust as necessary.

test/performance/user_test.rb
```diff
@@ -0,0 +1,42 @@
...
+    test 'equal to email validation is performant for valid passwords' do
+      user1 = ConfigOffUser.new(email: 'speed@aol.com', password: 'Password1!')
+      user1.valid?
+      user2 = ConfigOnUser.new(email: 'speed@aol.com', password: 'Password1!')
+      user2.valid?
+
+      result1 = Benchmark.measure { user1.valid? }
+      result2 = Benchmark.measure { user2.valid? }
+
+      assert_in_delta(result1.real, result2.real, 0.005)
```
dillonwelch, line 24: Ok, I'm finding that this still fails every so often, so I'm thinking to either bump the delta way up or just get rid of this. Any opinions?

olbrich: I'm not sure how useful these tests are. You're basically checking to see if you do things twice that they happen about as fast, which is what I would expect for almost anything.

lib/devise-security/models/secure_validatable.rb
olbrich: @dillonwelch is it possible to set the password to the email first and then update the record with the email and thus successfully bypass this check?

test/performance/user_test.rb
dillonwelch: This is because I want to make sure that the performance is the same regardless if the config is on or off. Before the change in this PR, the config being on resulted in this taking about 3x as long. 

Maybe this test is an over-reaction?

olbrich: Understood, and I have mixed feelings about it since that was the initial bug report.

lib/devise-security/models/secure_validatable.rb
dillonwelch: To make sure I understand, you're thinking something like this (skipping confirmation for brevity)?
```ruby
user = User.create(email: 'bob@aol.com', password: 'whatever')
user.update(password: 'jim@aol.com')
user.update(email: 'jim@aol.com')
```

test/performance/user_test.rb
dillonwelch: I can't think of a way to set this up without flakiness without a high enough delta that the test becomes meaningless. Unless you have any better ideas, I'll probably just remove it?

lib/devise-security/models/secure_validatable.rb
olbrich: Similar.  Maybe something like.
```
user = User.create(email: 'bob@aol.com', password: 'foo@example.com')
user.update(email: 'foo@example.com')
```

test/performance/user_test.rb
olbrich: I don't.  If there was enough need we could do a benchmark suite that would re-run the methods several thousand times to average out the noise, but I don't think that's the case normally.

lib/devise-security/models/secure_validatable.rb
dillonwelch: So I was able to test that we do not currently error out in this scenario. We don't currently run this validation if the user is not updating their password. This is because we wouldn't have the password in memory and so wouldn't be able to compare it to the email. I'm not sure of a way to get around this?

dillonwelch:
```ruby
user = User.create!(email: 'bob@aol.com', password: 'foO1@example.com', password_confirmation: 'foO1@example.com')
user = nil
user = User.first
user.email = 'foO1@example.com'

user.valid? # true
```

olbrich: We would have to encrypt the password the same way we do when we save it and compare the hashes, which might have been what it was originally doing.

dillonwelch: Maybe I'm missing something, but when I revert back to the original code I'm getting different hashes. 
```ruby
(byebug) email
"foo1@example.com"
(byebug) email_was
"bob@aol.com"
(byebug) encrypted_password
"$2a$12$PapCw7vqW6Tcr7t2ADrv3.yfDJm1Y8HevAexrJRZ4C.1L6y8rsjfu"
(byebug) dummy.encrypted_password
"$2a$12$PJRF1lYAXS9tLy/RCfCp/.TAO.EMLXzPUpVL/1C0XXKt3LA.mgtyW"
(byebug) dummy.password
"foo1@example.com"
(byebug) encrypted_password_was
"$2a$12$PapCw7vqW6Tcr7t2ADrv3.yfDJm1Y8HevAexrJRZ4C.1L6y8rsjfu"
(byebug) dummy
#<TestSecureValidatable::User id: nil, username: nil, facebook_token: nil, unique_session_id: nil, email: "", password_changed_at: nil, created_at: nil, updated_at: nil, paranoid_verification_code: nil, paranoid_verified_at: nil, paranoid_verification_attempt: 0, security_question_id: nil, security_question_answer: nil, expired_at: nil, last_activity_at: nil>

olbrich (approved):

PR #356, opened 2021-12-21
Per instance configs v2

dillonwelch:
Based on https://github.com/devise-security/devise-security/pull/63, but without the documentation comments (want to do this in a separate PR to make reviewing easier).


lib/devise-security/models/secure_validatable.rb
```diff
@@ -99,8 +113,30 @@ def email_required?
...
+        self.class.email_validation
+      end
+
+      def password_complexity
+        self.class.password_complexity
+      end
+
+      def password_length
+        self.class.password_length
+      end
```
olbrich, line 130: It may be possible to just delegate these calls to the class instead of defining new methods.

test/performance/user_test.rb
```diff
@@ -0,0 +1,42 @@
+require 'test_helper'
+
+require 'benchmark'
+
+module Performance
```
olbrich, line 5: I think these changes also exist in a different PR / Branch

dillonwelch: Yup, wanted the user speedup changes so that tests would run faste

lib/devise-security/models/secure_validatable.rb
dillonwelch: Updated to use this!

test/performance/user_test.rb
dillonwelch: If everything is good in this PR, I'll remove this code so that it can be merged

test/test_secure_validatable_overrides.rb
```diff
@@ -0,0 +1,154 @@
...
+      email: 'bob@microsoft.com',
+      password: 'Pa3zZ1!',
+      password_confirmation: 'Pa3zZ1!'
+    )
+
+    assert user.invalid?
+    assert_equal(
+      [
+        'Password is too short (minimum is 10 characters)',
+        'Password must contain at least 10 letters'
```
dillonwelch, line 132: Can you add separate tests for the complexity validators? I was trying to make it so that only one thing would fail in each test block.

olbrich: Sure!

lib/devise-security/models/secure_validatable.rb
```diff
@@ -46,14 +46,31 @@ def self.included(base)
...
           # extra validations
-          validates :email, email: email_validation if email_validation # see https://github.com/devise-security/devise-security/blob/master/README.md#e-mail-validation
-          validates :password,
-                    'devise_security/password_complexity': password_complexity,
-                    if: :password_required?
+          # see https://github.com/devise-security/devise-security/blob/master/README.md#e-mail-validation
+          validate do |record|
+            if email_validation
+              validates_with(
+                EmailValidator, { attributes: :email }
```
schlick, line 63: Ah, this change breaks the ability to pass in validation options. I have:
```
config.email_validation = { validate_mx: false }
```

PR #364, opened 2021-12-29
Update rubocop

olbrich:
* Removes version constraint on Rubocop
* Add rubocop-minitest

I deliberately did not apply any rubocop auto corrections in this branch. There are a lot of them and not all of them are correct or desirable, so we should do those in smaller batches.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,5 +1,6 @@
 require:
   - rubocop-rails
+  - rubocop-minitest
 
 AllCops:
   Exclude:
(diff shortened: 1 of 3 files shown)


.rubocop.yml
```diff
@@ -1,5 +1,6 @@
 require:
   - rubocop-rails
+  - rubocop-minitest
```
dillonwelch, line 3: What stuff does this bring in?

olbrich: Adds some additional Cops for minitest-specific things.

PR #367, opened 2021-12-29
Round 1 of Rubocop cleanup

dillonwelch:


olbrich:
You may want to merge #364 before you get too far along with updating. I'm sure some new cops will be introduced.

dillonwelch:
Is that one good to go?

dillonwelch:
@olbrich updated with the necessary changes from the minitest cops!

PR #369, opened 2021-12-29
Banned passwords

dillonwelch:
Rough draft for being able to ban passwords. Made this flexible so that we could have a module that allows you to add commonly used passwords, pwned passwords, and then any custom lists that an organization might went added to a ban list.

Open questions
- [ ] Do we want to add a `reason` field? (`common password`, `pwned`, etc)
- [ ] Do we want to add any sort of hook for caching banned password results?
- [ ] How do we want to add the list of common passwords (from https://github.com/danielmiessler/SecLists/tree/master/Passwords, but filtering out results of < 8 characters)
- [ ] Do we want to add a task to pull from pwned passwords (like https://github.com/michaelbanfield/devise-pwned_password/ but as a one-time task to update this table instead of each time we save a user)
- [ ] Do we want to add a config to check the password list each time a user interacts with the system (like in https://github.com/michaelbanfield/devise-pwned_password/)


dillonwelch:
Closing this as I want to look into other ways of doing this (bloom filter, calling pwned API directly, etc)

PR #371, opened 2022-01-04
Update to use byebug directly so we can use newer pry

dillonwelch:
Hoping this will fix build issue on ruby head


dillonwelch:
https://github.com/deivid-rodriguez/pry-byebug/pull/346 looks like there are some issues with the gem and latest versions of Ruby.

PR #401, opened 2022-08-09
Fix a grammer in a code comment

garigari-kun:
Hi

This is a small pull request to fix a grammer in a code comment.
Please check it out. Thanks!

### DIFF
--- a/lib/devise-security/models/password_archivable.rb
+++ b/lib/devise-security/models/password_archivable.rb
@@ -35,7 +35,7 @@ module Devise
         end
       end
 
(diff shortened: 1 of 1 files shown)


lib/devise-security/models/password_archivable.rb
```diff
@@ -35,7 +35,7 @@ def max_old_passwords
         end
       end
 
-      # validate is the password used in the past
+      # validate if the password used in the past
```
dillonwelch, line 38:
```suggestion
      # validate if the password was used in the past
```

PR #402, opened 2022-08-24
PasswordExpiredController: Accept block for update action

vlad-psh:
Currently if password is failed to update, nothing is displayed, and there is no way to change it easily (ie. without completely overriding `update` action in app's controller)

Devise though, allows user to add additional logic by passing a block to the `update`, like here: https://github.com/heartcombo/devise/blob/8593801130f2df94a50863b5db535c272b00efe1/app/controllers/devise/passwords_controller.rb#L34-L37

In this PR we're using same pattern.

So now user can do something like this:

```ruby
class Users::PasswordExpiredController < Devise::PasswordExpiredController
  def update
    super do |resource|
      flash[:error] = resource.errors.full_messages.first if resource.errors.any?
    end
  end
end
```

### DIFF
--- a/app/controllers/devise/password_expired_controller.rb
+++ b/app/controllers/devise/password_expired_controller.rb
@@ -20,7 +20,11 @@ class Devise::PasswordExpiredController < DeviseController
   # @see https://github.com/devise-security/devise-security/pull/111
   def update
     resource.extend(Devise::Models::DatabaseAuthenticatablePatch)
-    if resource.update_with_password(resource_params)
+    resource.update_with_password(resource_params)
+
+    yield resource if block_given?
+
+    if resource.errors.empty?
       warden.session(scope)['password_expired'] = false
       set_flash_message :notice, :updated
       bypass_sign_in resource, scope: scope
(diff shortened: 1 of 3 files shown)


dillonwelch:
As far as I'm aware, errors should be appearing on the page if the password fails to update. You can see a test that demonstrates this [here](https://github.com/devise-security/devise-security/blob/master/test/controllers/test_password_expired_controller.rb#L68). If you are still experiencing this issue, can you reproduce it in a brand new project? If so, can you create a dummy project and share it in example project here?

If you look at the commit that added the equivalent code to the `devise` project, you can see they added it for doing additional behavior such as enqueueing a job, not because errors didn't display.

I'm still happy to accept this PR for that reason, but I will need you to add a test for the behavior.

vlad-psh:
@dillonwelch Thanks for your comment!

Yeah, you're right, default device's view template renders list of `resource.errors`.

However, if rails app, for example, configured to display errors from flash messages right in the application layout, and we want to put our errors there, yielding block at `PasswordExpiredController` would be perfect solution for this case :)

PS: I've added new test similar to how it is implemented in `devise`

PR #414, opened 2023-03-31
Fix for Devise 4.9.1

### LINKED ISSUE #413: not working with the latest release of devise
in latest release of devise the method activerecord51? in Devise module was removed/renamed, thus breaking this gem

olivier-thatch:
The `Devise.activerecord51?` method was renamed to `Devise.active_record_51?` in Devise 4.9.1 (specifically [this commit](https://github.com/heartcombo/devise/pull/5576/commits/207ddc5127fcecbbcf5abcd8744775509e55b2e8)).

While I was at it, I updated the CI to start testing against modern Ruby and Rails versions, and fixed the CI badge in the README.

Fixes #413.


benjaminwols (approved):

lib/devise-security/models/compatibility/active_record_patch.rb
```diff
@@ -7,7 +7,11 @@ class NotPersistedError < ActiveRecord::ActiveRecordError; end
 
       module ActiveRecordPatch
         extend ActiveSupport::Concern
-        unless Devise.activerecord51?
+
+        active_record_51 = (Devise.respond_to?(:active_record_51?) && Devise.active_record_51?) ||
```
rocket-turtle, line 11: This check seems not to work:

(byebug) ActiveRecord.gem_version
#<Gem::Version "6.1.7.3">

### devise 4.8
> Devise.activerecord51?
true

### devise 4.9.1
> active_record_51 = (Devise.respond_to?(:active_record_51?) && Devise.active_record_51?) || (Devise.respond_to?(:activerecord51?) && Devise.activerecord51?)
false

The whole `unless Devise.activerecord51?` should be removed.
1. `saved_change_to_encrypted_password` seems not to be used
2. `encrypted_password_before_last_save` seems not to be used
3. `will_save_change_to_encrypted_password?` is used for Rails <= 5.0 which is out of maintanance for a long time.
4. If someone uses Rails 5.0 or older `devise-security` can be pinned to a working version.
5. `Devise.activerecord51` is not a public method.

simonbaird: Re "The whole unless Devise.activerecord51? should be removed.", #415 does this.

simonbaird:
See also #415 .

lib/devise-security/models/compatibility/active_record_patch.rb
olivier-thatch: 👍  Thanks. I've updated the PR to use the underlying implementation of `Device.active_record_51?`. I can also remove the entire patch if it's indeed unnecessary, but I'm not familiar enough with the codebase to tell for sure if that's the case.

cc @dillonwelch @olbrich if you have any recommendations

tarellel (approved):

olbrich:
Sorry for the delay in responding to this, I've been on vacation.

lib/devise-security/models/compatibility/active_record_patch.rb
olbrich: - Dropping old rails version support is outside the scope of this PR
- I think this is the only place where this method was used or needed, so this should be fine.

olbrich (approved):

olivier-thatch:
Thanks for the merge! Are you planning on cutting a new release soon?

olbrich:
@olivier-thatch Yes, I'm going through and making sure all the other stuff in master is ready to go as well, so hopefully this weekend.

olivier-thatch:
Wonderful, I was just able to bump devise and devise-security without issues. Thanks for the quick release :)

dillonwelch:
Thanks for reporting the issue + making a PR for it! Apologies I wasn't able to get to it either - I haven't been feeling so great.

smculliton:
Is there any way to backport this fix to v0.13.0? I could make a PR for a 0.13.0-stable version if it makes it easier

dillonwelch:
Unfortunately not. That version is very old, allows for a version of Devise that does not have this problem, and also requires a Rails version that is no longer supported and so falls outside of our maintenance policy.

PR #415, opened 2023-03-31
Fix compatibility with devise > 4.9.0

kivanio:
Remove dependence of internal method `Devise.activerecord51?`

Fixes 413


olbrich:
Closed in favor of #414

Issue #424, opened 2023-08-03
Why Validate Email Uniqueness?


MattKitmanLabs:
Why are we validating email uniqueness here?

https://github.com/devise-security/devise-security/blame/29ac90938436471498353328b61ee82d3e946341/lib/devise-security/models/secure_validatable.rb#L44

 If the email is not one of the authentication_keys, then maybe it should be down to the developers on each individual project to manage how they validate the email and within what scope.

Any thoughts or feedback on this?

dillonwelch:
Can you elaborate on why someone would want multiple users to share an email address?

We could add a `email_uniqueness_required?` method and set it to true by default, and then the dev could override as desired.

PR #425, opened 2023-08-19
Allow users to set config to skip email uniqueness validation.

MatthewKennedy:
Adds the ability to skip the email uniqueness validation and still use the other email validations.

**Use case:**
The setup uses the scope of `[:username, :store_id]` and the email is not unique, but still wants to use the other email validations provided by this gem, but not the uniqueness validation.


lib/devise-security/models/secure_validatable.rb
```diff
@@ -30,19 +30,17 @@ def self.included(base)
...
 
             already_validated_email = login_attribute.to_s == 'email'
           end
 
           unless devise_validation_enabled?
             validates :email, presence: true, if: :email_required?
-            unless already_validated_email
-              validates :email, uniqueness: true, allow_blank: true, if: :email_changed? # check uniq for email ever
-            end
+            validates :email, uniqueness: true, allow_blank: true, if: :email_changed? unless already_validated_email || skip_email_uniqueness_validation
```
dillonwelch, line 43: I think this should be a method `validate_email_uniquenes?` similar to the `email_required?` method.

README.md
dillonwelch, line 1: Can you separate out any README changes into a separate PR please? Easier to review that way :)

dillonwelch (changes requested):
Looks like this would take care of https://github.com/devise-security/devise-security/issues/424.

Make it a method vs another configuration option and I'm cool with it.

dillonwelch:
Thanks for the PR! Much appreciated.

dillonwelch:
I'm not sure - I inherited this codebase. It's probably not intentional :)

Did it cause test issues for you?

lib/devise-security/models/secure_validatable.rb
```diff
@@ -30,19 +30,17 @@ def self.included(base)
...
 
             already_validated_email = login_attribute.to_s == 'email'
           end
 
           unless devise_validation_enabled?
             validates :email, presence: true, if: :email_required?
-            unless already_validated_email
-              validates :email, uniqueness: true, allow_blank: true, if: :email_changed? # check uniq for email ever
-            end
+            validates :email, uniqueness: true, allow_blank: true, if: :email_changed? && :validate_email_uniquenes? unless already_validated_email
```
dillonwelch, line 43:
```suggestion
            validates :email, uniqueness: true, allow_blank: true, if: :email_changed? && :validate_email_uniqueness? unless already_validated_email
```

lib/devise-security/models/secure_validatable.rb
```diff
@@ -118,6 +116,10 @@ def email_required?
         true
       end
 
+      def validate_email_uniquenes?
```
dillonwelch, line 119:
```suggestion
      def validate_email_uniqueness?
```

test/test_secure_validatable_auth_key_username.rb
```diff
@@ -0,0 +1,55 @@
...
+    include ::Mongoid::Mappings if DEVISE_ORM == :mongoid
+
+    protected
+
+    def validate_email_uniquenes?
+      false
+    end
+  end
+
+  test "new user can use an existing user's email if validate_email_uniquenes? is set to false" do
```
dillonwelch, line 19:
```suggestion
  test "new user can use an existing user's email if validate_email_uniqueness? is set to false" do
```

test/test_secure_validatable_auth_key_username.rb
```diff
@@ -0,0 +1,55 @@
...
+class TestSecureValidatableAuthKeyUsername < ActiveSupport::TestCase
+  class User < ApplicationRecord
+    devise authentication_keys: [:username]
+
+    devise :database_authenticatable, :secure_validatable
+    include ::Mongoid::Mappings if DEVISE_ORM == :mongoid
+
+    protected
+
+    def validate_email_uniquenes?
```
dillonwelch, line 14:
```suggestion
    def validate_email_uniqueness?
```

dillonwelch (commented):
Typo 😅

dillonwelch (approved):

PR #439, opened 2024-01-09
Add `required_fields` models method

### LINKED ISSUE #436: Missing 'required_fields' method within devise extensions
There is built-in `Devise::Models.check_fields! Model` method which checks for required DB columns per enabled Devise extension for the Model class. Under the hood it tries to call 'required_fields' method for extension module (such as `Devise::Models::PasswordArchivable` or whatever).

At the moment it raises `NoMethodError` or deprecation warning (since https://github.com/heartcombo/devise/pull/1697) against `devise-security` extensions. Might be fixed like:

```ruby
module Devise
  module Models
    module PasswordArchivable
      def self.required_fields(klass)
        []
      end
    end
  end
end

```

ayerdines:
### Summary
This PR adds `self.required_fields(Model)` methods to all the models. `Devise::Models.check_fields! Model` calls this method under the hood to verify if the devise model has required columns present or not.

fixes #436

### DIFF
--- a/lib/devise-security/models/secure_validatable.rb
+++ b/lib/devise-security/models/secure_validatable.rb
@@ -82,6 +82,10 @@ module Devise
         raise "Could not use SecureValidatable on #{base}" unless base.respond_to?(:validates)
       end
 
+      def self.required_fields(_klass)
+        []
+      end
+
       def current_equal_password_validation
         return if new_record? || !will_save_change_to_encrypted_password? || password.blank?
 
--- a/.github/workflows/test_suite.yml
+++ b/.github/workflows/test_suite.yml
@@ -13,7 +13,7 @@ jobs:
       matrix:
         orm: [active_record, mongoid]
         rails: ["5.2", "6.0", "6.1", "7.0"]
-        ruby: ["2.7", "3.0", "3.1", "3.2", head]
+        ruby: ["2.7", "3.0", "3.1", "3.2", "3.3", head]
         exclude:
           - rails: 5.2
             ruby: 3.0
@@ -21,6 +21,8 @@ jobs:
             ruby: 3.1
           - rails: 5.2
             ruby: 3.2
+          - rails: 5.2
+            ruby: 3.3
           - rails: 5.2
             ruby: head
           - rails: 7.0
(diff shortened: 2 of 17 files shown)


lib/devise-security/models/secure_validatable.rb
```diff
@@ -82,6 +82,10 @@ def self.assert_secure_validations_api!(base)
         raise "Could not use SecureValidatable on #{base}" unless base.respond_to?(:validates)
       end
 
+      def self.required_fields(_klass)
+        []
```
dillonwelch, line 86: Seems like this should use `klass` to conditionally require based on the validations starting on line 25?

dillonwelch: Example from Devise: https://github.com/heartcombo/devise/blob/main/lib/devise/models/confirmable.rb#L72

ayerdines: Thank you so much for the review @dillonwelch ❤️ 

I think that `self.required_fields` should only include the attributes that are absent in the devise model table by default and required by this module. Since the `email` and `password` fields are added by the default devise migration,  we should be okay to keep this empty. It's empty for the `Devise::Models::Validatable` as well https://github.com/heartcombo/devise/blob/e2242a95f3bb2e68ec0e9a064238ff7af6429545/lib/devise/models/validatable.rb#L22.
Please let me know if I'm missing anything @dillonwelch. 🙏

ayerdines: Also, the tests for the ruby `head` version i.e. `3.4.0` are failing because of the dependency of `minitest` on `mutex_m` which has been moved from default gems to bundled gems. Should I fix it in this PR, I'm thinking about adding the following to `gemspec` file.
```ruby
s.add_development_dependency 'mutex_m'
```

dillonwelch: Required fields: ok, that makes sense

Gem issue: Yes please!!!!!

ayerdines:
@dillonwelch great!!
I have pushed a fix. Please help me review this again.

Thank you!! :)

dillonwelch (approved):

PR #440, opened 2024-01-26
fix: specify attribute name on custom validator

pdkproitf:


dillonwelch:
Looks like your changes blew up the test suite. Can you look into this?

pdkproitf:
> Looks like your changes blew up the test suite. Can you look into this?

thank you, I'm working on it.

pdkproitf:
Due to the code change, I found a problem in test case `password complexity can be overridden at the class level`

At point 1, when `secure_validatable` is specified, the `User` class registers a validation with the default password validator. Subsequently, the `password_complexity_validator` value at point 2 does not have an effect unless point 2 is declared before point 1.

Could you give some suggestions?
cc: @dillonwelch 

Screenshot_2024-01-31_at_14_45_00

dillonwelch:
Looking at the core code in `activemodel`, it looks like the difference is that [validate](https://github.com/rails/rails/blob/main/activemodel/lib/active_model/validations.rb#L171) actually sets a callback instead of registering a validator like [validates_with](https://github.com/rails/rails/blob/main/activemodel/lib/active_model/validations/with.rb#L88) does. 

I am not sure yet what the right option / fix here is yet. I will continue to think on this. Hopefully this gives you a new avenue to explore to fix it. I do still agree that you should be able to pull in these validators when you do `User.validators_on(:password).map(&:class)`.

dillonwelch:
The tests that were broken on main should be fixed now, hopefully that helps with troubleshooting.

PR #442, opened 2024-02-10
feat: Support Traceable Sessions

itsmechlark:
Allows having a history of login sessions with flexible active session limits.

Related #71


dillonwelch (commented):
Looks like merge conflicts now since you so kindly fixed the test suite in a separate PR.

Can you look at some of the changes you made to the test config and see if they could be pulled out separately as well? Would make the review on this one easier, which I want to be thorough on since it's a new module and impacts an existing module. Also, I want to get other changes in so I can cut a new minor version and then have this in a new major version.

Note that I'm reviewing this while very sleepy and will review again in the near future with a more clear mind.

lib/devise-security/models/session_traceable.rb
```diff
@@ -0,0 +1,111 @@
+# frozen_string_literal: true
+
+require_relative "#{DEVISE_ORM}/session_history"
+require 'devise-security/hooks/session_traceable'
+
+module Devise
+  module Models
+    # SessionTraceable takes care of session trail in every authentication.
+    # When a session expires after expiring the +token+, the user
+    # will be asked for credentials again, it means, he/she will be redirected
```
dillonwelch, line 10: This sentence feels like it runs on too long.

lib/devise-security/models/session_limitable.rb
```diff
@@ -31,12 +31,71 @@ def update_unique_session_id!(unique_session_id)
...
+          next unless session.last_accessed_at && timeout_in && session.last_accessed_at <= timeout_in.ago
+
+          session.update_attribute_without_validatons_or_callbacks(:active, false)
+        end
+      end
+
+      # Maximum number of active sessions
+      # @return [Numeric]
+      # @return [1] by default. This can be overridden by application logic as necessary.
+      def max_active_sessions
```
dillonwelch, line 66: After reading through this again, it's unclear to me why we would want to allow more than one session in a module designed specifically to limit to one session.

lib/devise-security/models/session_limitable.rb
dillonwelch, line 1: Reading through this again, this feels separate from the `SessionTraceable` module you're adding. 

If this is the case, it feels like it would be better as a separate PR.

If this is not the case, please let me know.

lib/devise-security/models/session_traceable.rb
```diff
@@ -0,0 +1,111 @@
...
+      def self.required_fields(_klass)
+        %i[session_history_class session_ip_verification]
+      end
+
+      # Log user session.
+      # @return [true] if session is allowed to log.
+      # @return [false] if session isn't allowed to log when already exceeded to `:max_active_sessions`
+      #   or session is not valid.
+      def log_traceable_session!(options = {})
+        token = options[:token] || generate_traceable_token
```
dillonwelch, line 30: Wondering if we should call this `session_history_token` instead of just `token`, to make it clear that it's not a token for the session but for its history.

lib/devise-security/models/session_traceable.rb
```diff
@@ -0,0 +1,111 @@
...
+        opts = options.merge(token: token, last_accessed_at: Time.current.utc)
+        opts = session_traceable_condition(opts)
+
+        return false if devise_modules.include?(:session_limitable) && !allow_limitable_authentication?
```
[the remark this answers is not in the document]
dillonwelch: Also, I am not sure why this is here - I think it would be OK to have session tracing without having session limiting.

lib/devise-security/models/session_traceable.rb
```diff
@@ -0,0 +1,111 @@
...
+
+        session_traceable_adapter.create!(opts) && token
```
dillonwelch, line 36: Not sure why the `&& token` is here, won't that always just be truthy?

lib/devise-security/models/session_traceable.rb
dillonwelch, line 1: More thorough documentation on the individual methods, including params and any side effects, would be appreciated.

lib/devise-security/models/session_traceable.rb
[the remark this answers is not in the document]
dillonwelch: Unless the intent is only to allow this if session limiting is also in place?

lib/devise-security/models/mongoid/session_history.rb
dillonwelch, line 1: Given that this is supposed to be a history record, it seems like it would be more granular to create a new record each time the session was accessed instead of having one record for the session with `last_accessed_at` being updated.

dillonwelch (commented):
Pressed the button before I was done :)

lib/devise-security/hooks/session_traceable.rb
```diff
@@ -0,0 +1,65 @@
...
+
+# Each time a record is fetched from session we verify if the session
+# has a valid unique session identifier. If so, then we set the last accessed time.
+Warden::Manager.after_set_user only: :fetch do |record, warden, options|
+  scope = options[:scope]
+  env = warden.request.env
+  opts = { ip_address: warden.request.remote_ip }
+  if record.devise_modules.include?(:session_traceable) &&
+     warden.authenticated?(scope) &&
+     options[:store] != false &&
```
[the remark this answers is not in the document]
itsmechlark: By default, `store` is always a `NilClass` unless `warden.authenticate` is called with an option `store: false` or on `devise` calling `authenticate_user` with `store: false` so it will be safer to use that condition which follows the [convention](https://github.com/heartcombo/devise/blob/1d6658097e364d45b5e059976f1e822eee7d67da/lib/devise/hooks/rememberable.rb#L5) in `devise`.

lib/devise-security/models/mongoid/session_history.rb
itsmechlark: If we create a session history every time a user is fetch (on `warden.authenticate` or on `devise` `current_user`) then it will create a lot of session histories. Let's say in a single session of a user every time rails controller is hit, a history which likely in a single session it will be thousand of requests.

The main purpose of `last_accessed_at` is to track when the specific session (which track using token) is being used. The data relationship will always be a user has many sessions and last activity is tracked.

lib/devise-security/models/session_limitable.rb
itsmechlark: This changes will allow to maintain the original logic of `session_limitable` which only allow 1 session at a time when `session_traceable` module but when `session_traceable` is added then it add flexibility on session limitable to customize maximum sessions per user and allowing session to expire. On scenarios that a user already hit the limit then its either reject authentication or expire the oldest session.

lib/devise-security/models/session_limitable.rb
```diff
@@ -31,12 +31,71 @@ def update_unique_session_id!(unique_session_id)
         end
       end
 
+      # Removes old/inactive session if allowed.
```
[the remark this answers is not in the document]
itsmechlark: The purpose of `allow_limitable_authentication?` is to:
1. If `session_traceable` module is not included then it will always be `true`.
2. If didn't matched to 1 and number of sessions is still within the allowable number of session then it will allow authentication.
3. If didn't matched to 2 and will only reject authentication on limit then it will try to remove any expired session to accomodate new authentication using `deactivate_timeout_sessions!`.
4. If didn't matched to 3 (which reject authentication is disabled) then it will expire the oldest session regardless if its already expired.
5. If didn't matched to 4 then disallow authentication.

lib/devise-security/models/session_limitable.rb
```diff
@@ -31,12 +31,71 @@ def update_unique_session_id!(unique_session_id)
...
+      def allow_limitable_authentication?
+        # Always fallback to default logic when session_traceable is not supported.
+        return true unless devise_modules.include?(:session_traceable)
+
+        opts = session_traceable_condition(active: true)
+        return true if max_active_sessions > session_traceable_adapter.find_all(opts).size
```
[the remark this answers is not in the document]
itsmechlark: `session_traceable_adapter` will always be the `session_history_class` ORM compatible on `session_traceable` module. It will count the number of active sessions to check whether it is already on its maximum allowable number of sessions.

lib/devise-security/models/session_limitable.rb
```diff
@@ -31,12 +31,71 @@ def update_unique_session_id!(unique_session_id)
...
+
+        opts[:order] = %i[last_accessed_at asc]
+        session = session_traceable_adapter.find_first(opts)
+        return false unless session
+
+        session.update_attribute_without_validatons_or_callbacks(:active, false)
+      end
+
+      # Deactivate expired sessions
+      def deactivate_timeout_sessions!
```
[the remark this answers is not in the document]
itsmechlark: Agree!

lib/devise-security/models/session_limitable.rb
itsmechlark: Aside from what's mentioned [here](https://github.com/devise-security/devise-security/pull/442/files#r1490438209), `session_traceable` can work without `session_limitable` module, and same with `session_limitable`. This give more flexibility to the number of active sessions allowed in every user which still put limits.

lib/devise-security/models/session_limitable.rb
```diff
@@ -31,12 +31,71 @@ def update_unique_session_id!(unique_session_id)
...
+      # @return [Numeric]
+      # @return [1] by default. This can be overridden by application logic as necessary.
+      def max_active_sessions
+        # Always fallback to default logic when session_traceable is not supported.
+        return 1 unless devise_modules.include?(:session_traceable)
+
+        self.class.max_active_sessions
+      end
+
+      # Reject session when exceeded to allowed number of active sessions
```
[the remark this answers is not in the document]
itsmechlark: What do you think of the following:
```
# Always rejects sessions when already exceeded to allowable number of sessions.
# @return [Boolean]
# @return [false] by default when `session_traceable` is not supported and use `reject_sessions` configuration when not. This can be overridden by application logic as necessary.
```

lib/devise-security/models/session_traceable.rb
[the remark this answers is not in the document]
itsmechlark: Agree, it should happen first. And yes, when  `session_limitable` module is enabled, the intent is log session when allowed by session limitable.

lib/devise-security/models/session_traceable.rb
itsmechlark: It will ensure that we return the token when we are able to log it.

lib/generators/active_record/devise_security_generator.rb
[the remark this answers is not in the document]
itsmechlark: Sure.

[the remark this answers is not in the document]
itsmechlark: Please see PR #447.

dillonwelch:
Thank you for the quick responses. I will get to them soon.

lib/devise-security/models/session_traceable.rb
itsmechlark: I update the docs to make it more clearer what the `token` is.

lib/devise-security/models/session_limitable.rb
```diff
@@ -31,12 +31,71 @@ def update_unique_session_id!(unique_session_id)
...
+          session.update_attribute_without_validatons_or_callbacks(:active, false)
+        end
+      end
+
+      # Maximum number of active sessions
+      # @return [Numeric]
+      # @return [1] by default. This can be overridden by application logic as necessary.
+      def max_active_sessions
+        # Always fallback to default logic when session_traceable is not supported.
+        return 1 unless devise_modules.include?(:session_traceable)
```
dillonwelch, line 68: Why do we need to have `session_traceable` turned on before allowing more than 1 active session?

lib/devise-security/models/session_limitable.rb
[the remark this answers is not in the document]
dillonwelch: Unclear to me why we can't just use `SessionHistory`.

lib/devise-security/models/session_limitable.rb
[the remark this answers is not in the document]
dillonwelch: Steps 3-5 don't feel like they belong in this method - boolean methods should only be returning true / false based on whatever logic you're adding. Steps 1-2 seem fine and what this method should be used for.

dillonwelch (commented):
Can fix the merge conflicts now since I merged in the other PR.

PR #443, opened 2024-02-13
ci: Github Codespaces configuration

itsmechlark:
### Summary

Github Codespaces allows (new) `devise-security` contributors to easily boot a fully functional environment to make and test changes to `devise-security` on any ruby versions. In order for this repository to work will with Codespaces, we need some configuration, this is added in this PR.

The `Dockerfile` and `boot.sh` are inspired on the [Development Dependencies Install manual](https://edgeguides.rubyonrails.org/development_dependencies_install.html) and [devise_auth0 devcontainer](https://github.com/itsmechlark/devise_auth0/tree/main/.devcontainer).


Gemfile
```diff
@@ -8,6 +8,7 @@ gem 'minitest-rails', '~> 6.0.0'
 gem 'railties', '~> 6.0.0'
 
 group :active_record do
+  gem 'pg'
```
dillonwelch, line 11: Wonder if adding this is what's causing the tests to fail (in the other PR too?)

.devcontainer/boot.sh
```diff
@@ -0,0 +1,9 @@
+sudo chown -R vscode:vscode ${BUNDLE_PATH} && \
+    bundle install --jobs=3 --retry=3 && \
+    bundle exec appraisal install
+
+# Rebuild PostgreSQL
+dropdb --if-exists devise_security_test
```
dillonwelch, line 6: Can you use the rails commands here instead (not sure)?

dillonwelch:
https://github.com/devise-security/devise-security/pull/440 wondering if the failure here is related / if there's a failure in main???

itsmechlark:
> #440 wondering if the failure here is related / if there's a failure in main???

Hi @dillonwelch! Yes, looks like there's a change somewhere that affects `database_cleaner` setup on minitest lifecycle which start and clean the database isn't called.

Gemfile
itsmechlark: No, it is more related to database_cleaner I guess. On #442, I fixed it by generating unique email for those test cases that uses same email address for users.

.devcontainer/boot.sh
itsmechlark: We can but it will add more complexity as `boot.sh` uses PostgreSQL builtin CLI to rebuild the database regardless it is DevContainer/Codespaces or host development.

dillonwelch: Ok, this is fine then

dillonwelch (approved):

PR #444, opened 2024-02-14
test: Fixes failing tests due to not unique email

itsmechlark:


test/support/helpers.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+# Copyright 2009-2019 Plataformatec. http://plataformatec.com.br
```
dillonwelch, line 3: Can you remove this copyright bit?

dillonwelch:
Can you take a look at the mongo failure if it's not too much trouble? Looks like a require statement change?

itsmechlark:
> Can you take a look at the mongo failure if it's not too much trouble? Looks like a require statement change?

Done, added `bigdecimal` and `drb` to gems since it no longer part of the default gems for Ruby 3.4.

dillonwelch (approved):
Thanks for working through these issues. You rock!

PR #452, opened 2024-10-21
fix: clear unique_session_id when user signs out

baueric:
Prevents session replays when application uses a cookie session store.


reinaris:
@baueric thanks for the effort. Hopefully this will be merged soon.

dillonwelch:
Thank you for your fix.

PR #455, opened 2024-12-06
Enable lazy loading of archivable configurations

FranciscoJRSantos:
Configurations `password_archiving_count` and `deny_old_passwords` now accept `Proc.new` or `lambda` invocations. This enables the configurations to be loaded during runtime when needed.


olbrich (changes requested):
@FranciscoJRSantos Thank you for your contribution!  Can you please add a couple of tests for this?

PR #457, opened 2024-12-27
Add Ruby 3.4 support & enable Rails 7.2 tests

olbrich:
- Add Ruby 3.4 to test matrix
- Enable Rails 7.2 testing now that minitest-rails has a 7.2 version
- Install RMagick in builds to allow tests to run


dillonwelch:
Yay!!!

PR #468, opened 2025-03-08
Fix i18n tests

dillonwelch:


dillonwelch:
Issue was with libs being removed from latest Ruby, see:
https://stackoverflow.com/questions/79360526/uninitialized-constant-activesupportloggerthreadsafelevellogger-nameerror
https://github.com/mongodb/mongoid/pull/3370

